PR #29, opened 2024-01-01
config fixes: import legacy, save as JSON

### LINKED ISSUE #6: Pre-4.0.0 Config Files Cannot Be Used and Need Migrated
Attempting to run the binary with `-h|--help` reports the following error when the program hasn't been run at least once:

`./dist/LogosLinuxInstaller -h`

```
ERROR:root:Error decoding config file /home/thwright/.config/Logos_on_Linux/Logos_on_Linux.conf: Expecting value: line 1 column 1 (char 0)
  File "LogosLinuxInstaller.py", line 241, in <module>
  File "LogosLinuxInstaller.py", line 146, in main
    switch_logging()
  File "config.py", line 104, in set_config_env
    logos_error(f"Error: Unable to get config at {config_file_path}")
NameError: name 'logos_error' is not defined
[11339] Failed to execute script 'LogosLinuxInstaller' due to unhandled exception!
```

This also applies to running the main script:

`./LogosLinuxInstaller.py -h`

```
ERROR:root:Error decoding config file /home/thwright/.config/Logos_on_Linux/Logos_on_Linux.conf: Expecting value: line 1 column 1 (char 0)
  File "/home/thwright/Dev/LogosLinuxInstaller/./LogosLinuxInstaller.py", line 223, in <module>
    main()
  File "/home/thwright/Dev/LogosLinuxInstaller/./LogosLinuxInstaller.py", line 128, in main
    config.set_config_env(config.CONFIG_FILE)
  File "/home/thwright/Dev/LogosLinuxInstaller/config.py", line 104, in set_config_env
    logos_error(f"Error: Unable to get config at {config_file_path}")
    ^^^^^^^^^^^
NameError: name 'logos_error' is not defined
```

n8marti:
Rather than log an error and exit if the config file can't be decoded as JSON, assume that it is a plaintext legacy config file. Read it line-by-line, assuming the key/value separator is '='.

I added an if statement (L112-L114) to remove any comments found *after* the `key=value` text, like `key=value # comment`, but maybe that's an unlikely (or even impossible) scenario?

This doesn't fully close #6, since there is a bigger discussion going on there, but it would at least remove the roadblock to proper script execution.

### DIFF
--- a/LogosLinuxInstaller.py
+++ b/LogosLinuxInstaller.py
@@ -119,6 +119,9 @@ def restore():
 def main():
     cli_args = parse_command_line()
 
+    # Initialize logging.
+    initialize_logging(logging.WARNING)
+
     die_if_running()
     die_if_root()
 
@@ -132,16 +135,16 @@ def main():
         config_file = config.LEGACY_CONFIG_FILE
     config.set_config_env(config_file)
 
+    # Parse CLI args and update affected config vars.
+    parse_args(cli_args)
+
     # Update logging level from config.
     logger = logging.getLogger()
     logger.setLevel(config.LOG_LEVEL)
 
-    parse_args(cli_args)
-
     if config.DELETE_INSTALL_LOG and os.path.isfile(config.LOGOS_LOG):
         os.remove(config.LOGOS_LOG)
-    initialize_logging(config.LOG_LEVEL)
-
+    
     # If Logos app is installed, run the desired Logos action.
     if config.LOGOS_EXE is not None and os.access(config.LOGOS_EXE, os.X_OK):
         if config.ACTION == 'control':
--- a/LogosLinuxInstaller.py
+++ b/LogosLinuxInstaller.py
@@ -147,10 +147,13 @@ def main():
     config.get_env_config()
 
     if config.DELETE_INSTALL_LOG and os.path.isfile(config.LOGOS_LOG):
-        os.remove(config.LOGOS_LOG)
+        # Write empty file before continuing.
+        with open(config.LOGOS_LOG, 'w') as f:
+            f.write('')
 
     # If Logos app is installed, run the desired Logos action.
     if config.LOGOS_EXE is not None and os.access(config.LOGOS_EXE, os.X_OK):
+        logging.info(f"App is installed: {config.LOGOS_EXE}")
         if config.ACTION == 'control':
             run_control_panel()
             sys.exit(0)
--- a/LogosLinuxInstaller.py
+++ b/LogosLinuxInstaller.py
@@ -26,6 +26,7 @@ from utils import getDialog
 from utils import set_appimage
 from utils import set_default_config
 from utils import setDebug
+from utils import write_config
 from wine import run_indexing
 from wine import run_logos
 from wine import run_winetricks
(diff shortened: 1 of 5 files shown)


thw26:
I had considered your .json proposal and I was thinking it'd be the best solution for us, just because it is easiest. We wouldn't be able to port over someone's previous install though and some kind of migration code would then be needed.

One solution would be to "find" all the vars ourselves rather than trying to convert them. One thought would be: have the user provide their current install dir and then we perform some commands to find what we need to fill in a new json config.

n8marti:
Would it work to:
1. define something like `LEGACY_CONFIG_FILE` pointing to `~/.config/Logos_on_Linux/Logos_on_Linux.conf`,
2. then, if `~/.config/Logos_on_Linux/Logos_on_Linux.json` doesn't exist, read in the config from the `.conf` file when the app starts. Any variables set there should I guess correspond to the most recently-installed Logos app and wine prefix, and so 
3. our app would read that legacy config just like it does the new JSON config in order to run the currently-installed Logos app, and then
4. it would also re-save the variables to the JSON file, which would take precedence going forward.

Does that describe the same process that you mean by "migration code"?

thw26:
> Would it work to:
> 
>     1. define something like `LEGACY_CONFIG_FILE` pointing to `~/.config/Logos_on_Linux/Logos_on_Linux.conf`,
> 
>     2. then, if `~/.config/Logos_on_Linux/Logos_on_Linux.json` doesn't exist, read in the config from the `.conf` file when the app starts. Any variables set there should I guess correspond to the most recently-installed Logos app and wine prefix, and so
> 
>     3. our app would read that legacy config just like it does the new JSON config in order to run the currently-installed Logos app, and then
> 
>     4. it would also re-save the variables to the JSON file, which would take precedence going forward.
> 
> 
> Does that describe the same process that you mean by "migration code"?

That sound like you give me too much credit, but that process sounds like a good solution!

I think it's quite possible that someone will boot the Python app expecting to use their previous install. One notorious Logos on Linux dev is still using Logos 25.2 from May 23 as his primary install because upgrades to Logos frighten him, even though he trusts in the sovereignty and providence of God.

If we implement this as you described it and as I dreamed it, I think it'd be a great upgrade over the bash script, which assumed Logos was not installed when run. LEGACY_CONFIG sounds like a great way to do this.

I'm not sure which var names changed, as many of them are still bash style (all caps) rather than python style (lower case). I think we most changed function names, right?

n8marti:
Right. All the variables that were formerly bash env variables retained the ALL_CAPS casing and I didn't change any of the names. They still function like global variables, which are commonly defined with all-caps, so I didn't see any need to change them. Any variables that I added for use within functions or classes are all lowercase.

I didn't change any function names after the initial port to python. However, any function I've added has the format `name_of_function` rather than `nameOfFunction`, so there is some mixing and matching in the current code.

n8marti:
okay, this is ready for review now. The main goal of implementing a config migration is met, as well as a couple of other kind-of-related tweaks (arparse wording, logging setup). Better to merge this sooner than later without any further changes to prevent "mission creep".

thw26:
Thanks for this. I will work on reviewing this in a few hours.

thw26:
Tested the config conversion. This works.

I also tested having no config. All appears to work well there, too.

Anything else to test?

n8marti:
No, I think that covers it

PR #31, opened 2024-01-02
Add an Option to Redownload Winetricks

### LINKED ISSUE #4: Add an option to reinstall winetricks
To get to beta, we need to add an option to reinstall winetricks in case the user deletes it. This is also necessary for something like #3.

- [ ] optarg
- [ ] CLI menu
- [ ] GUI menu

thw26:
This commit fixes #4.

It also places the TUI app in a while loop and moves the quick launch code behind an `if GUI` check.

Depends on #29.

### DIFF
--- a/LogosLinuxInstaller.py
+++ b/LogosLinuxInstaller.py
@@ -12,6 +12,7 @@ from control import open_config_file
 from control import remove_all_index_files
 from control import remove_library_catalog
 from installer import install
+from installer import setWinetricks
 from msg import cli_msg
 from msg import initialize_logging
 from msg import logos_error
@@ -56,6 +57,8 @@ def parse_command_line():
     parser.add_argument('--shortcut', '-s', action='store_true', help='Create shortcut')
     parser.add_argument('--passive', '-P', action='store_true', help='Install Faithlife product non-interactively')
     parser.add_argument('--control-panel', '-C', action='store_true', help='Open Control Panel app')
+    parser.add_argument('--get-winetricks', action='store_true', help='Download or update Winetricks')
+    parser.add_argument('--run-winetricks', action='store_true', help='Download or update Winetricks')
     return parser.parse_args()
 
 def parse_args(args):
@@ -79,6 +82,14 @@ def parse_args(args):
 
     if args.reinstall_dependencies:
         config.REINSTALL_DEPENDENCIES = True
+
+    if args.get_winetricks: 
+       setWinetricks()
+       sys.exit(0)
+
+    if args.run_winetricks:
+        run_winetricks()
+        sys.exit(0)
 
     if args.debug:
         setDebug()
@@ -151,22 +162,6 @@ def main():
         with open(config.LOGOS_LOG, 'w') as f:
             f.write('')
 
-    # If Logos app is installed, run the desired Logos action.
-    if config.LOGOS_EXE is not None and os.access(config.LOGOS_EXE, os.X_OK):
-        logging.info(f"App is installed: {config.LOGOS_EXE}")
-        if config.ACTION == 'control':
-            run_control_panel()
-            sys.exit(0)
-        elif config.ACTION == 'indexing':
-            run_indexing()
-            sys.exit(0)
-        elif config.ACTION == 'logging':
-            switch_logging()
-            sys.exit(0)
-        elif config.ACTION == 'app':
-            run_logos()
-            sys.exit(0)
-
     # Check for environment variables.
     if config.DIALOG is None:
         getDialog()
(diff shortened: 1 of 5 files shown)


LogosLinuxInstaller.py
```diff
@@ -163,58 +164,76 @@ def main():
...
-    else:
-        logos_error("Unknown menu choice.")
+        # If Logos app is installed, run the desired Logos action.
+        if config.LOGOS_EXE is not None and os.access(config.LOGOS_EXE, os.X_OK):
+            if config.ACTION == 'control':
+                run_control_panel()
+                sys.exit(0)
+            elif config.ACTION == 'indexing':
+                run_indexing()
+                sys.exit(0)
```
n8marti, line 174: These installed-app actions also get run as CLI options, so they're not GUI-only. For example, if a user has run the binary once and Logos gets installed, then they run it again with `LogosLinuxInstaller --run-winetricks`, it looks like they will still end up in the TUI menu rather than have winetricks start.

LogosLinuxInstaller.py
```diff
@@ -163,58 +164,76 @@ def main():
...
+                if config.LOGS == "DISABLED":
+                    options_installed.append("Enable Logging")
+                else:
+                    options_installed.append("Disable Logging")
+                options = options_default + options_installed + options_exit
+            else:
+                options = options_default + options_exit
+
+            choice = None
+            if config.DIALOG is None or config.DIALOG == 'tk':
```
n8marti, line 197: The GUI installer would start at this point, but since it falls inside the `config.GUI is False` block, it never starts.

LogosLinuxInstaller.py
```diff
@@ -163,58 +164,76 @@ def main():
...
+                sys.exit(0)
+            elif config.ACTION == 'app':
+                run_logos()
+                sys.exit(0)
+
+    if config.GUI is False:
+        while True:
+            options_default = ["Install Logos Bible Software"]
+            options_exit = ["Exit"]
+            if file_exists(config.CONFIG_FILE):
```
n8marti, line 186: This needs to updated to check for the LOGOS_EXE binary rather than the config file, because in the case of a deleted installation the config file still exists but none of the `options_installed` are actually runnable.

n8marti:
I need to figure out how to place my comments so the correct lines are previewed with them. Sorry about the mismatches above. They're easier to follow if you open the "Files Changed" tab.

Maybe we need to think through the logic of `main` a little bit more at a bigger-picture level. The CLI options in some cases preselect choices that would be found later in the TUI or GUI (e.g. `SKIP_FONTS`), but in other cases they completely override the need for the UI (e.g. `--run-winetricks`). This is why I added the `ACTIONS` variable. It served to short-circuit the UIs when this kind of non-interface "action" was requested from the CLI. So my knee-jerk reaction is to say "just put it back", but then that would mess up being able to put the TUI menu into a loop, which is a good idea.

But maybe the `if config.DIALOG is None or config.DIALOG == 'tk':` block (L197-L201) could be moved up and out of the `if config.GUI is False` block, along with the "actions" block (L167-L180), then the `if config.GUI is True|False` parts could be reverted while the `curses` block could just start with:
``
`if config.DIALOG == 'curses':
    while True:
        [...]
```
This was an `elif` statement originally. Would something like that work?

**Comment by thw26** [comment]
Thanks for the review. I will rebase and then try to fix up what you've shared.

**Comment by thw26** [comment]
Rebased. Now need to review and fix up your comments.

File: LogosLinuxInstaller.py (hunk 3 above)
Comment (thw26) [reply LogosLinuxInstaller.py:190]: Good catch. Fixed in a commit

File: LogosLinuxInstaller.py (hunk 2 above)
Comment (thw26) [reply LogosLinuxInstaller.py:201]: Moved it to the `if` block above.

File: LogosLinuxInstaller.py (hunk 1 above)
Comment (thw26) [reply LogosLinuxInstaller.py:178]: Fixed this by properly setting the optargs above

**Comment by thw26** [comment]
@n8marti, this should be ready for a new review.

**Comment by thw26** [comment]
Also, perhaps we can do a voice style to rethink the optargs/logic of main. I will open a discussion for this.

File: LogosLinuxInstaller.py [hunk 4]
```diff
@@ -80,6 +83,14 @@ def parse_args(args):
     if args.reinstall_dependencies:
         config.REINSTALL_DEPENDENCIES = True
 
+    if args.get_winetricks: 
+       setWinetricks
+       sys.exit(0)
```
n8marti, line 88: function needs to be called with `setWinetricks()`

LogosLinuxInstaller.py
```diff
@@ -80,6 +83,14 @@ def parse_args(args):
...
+
+    if args.run_winetricks:
+        run_winetricks()
+        sys.exit(0)
+
```
n8marti, line 93: If these functions are run here in `parse_args()`, then the config variables are not yet properly set when these functions run. This ties into discussion #34. At the point where `parse_args()` is called, the config file has been read, but other CLI options haven't yet been handled (e.g. `-V`) and the environment variables have not yet been read. This happens in this order so that environment variables take precedence over CLI options, which take precedence over the config file, which takes precedence over default config. So the CLI arguments have to be read and parsed, then the environment variables need to be read and parsed, then the runtime config can finally bet set, and only then can any "actions" run under the intended runtime config.

n8marti:
I think this PR is dependent on the outcome of #34, unless the scope is reduced to only add the new `--get-winetricks` and `--run-winetricks` options but not yet implement them.

LogosLinuxInstaller.py
thw26: Fixed in current draft, but per our discussion, this should probably set a var that we later evaluate.

LogosLinuxInstaller.py
thw26: See discussion

thw26:
> I think this PR is dependent on the outcome of #34, unless the scope is reduced to only add the new `--get-winetricks` and `--run-winetricks` options but not yet implement them.

Yes, agreed. I will push the syntax fix, and then we can discuss how we want to split optargs that are for runtime config and optargs that are actionable.

thw26:
Squashed and pushed these modifications.

I will merge this change, then we can merge the reinstall dependencies code, then we can discuss how best to handle the variable optargs and the action optargs.

PR #33, opened 2024-01-02
Add an Option to Reinstall Dependencies

### LINKED ISSUE #3: Add an option to reinstall depdencies to post-install menu
This is functionality originally intended for the 3.8.0 Bash update. @jg00dman has already determined this and it should be part of the script now, but we need the functionality to be accessible by the user.

This is particularly necessary for SteamOS.

Depends on #1.

thw26:
Fixes #3.

The install routine for dependencies was broken. I had to test this heavily on my pacman machine but I got it working.

This PR thus fixes the install dependencies routine and adds an optarg, a TUI, and a GUI to run the check_dependencies() function.

Depends on #31.

### DIFF
--- a/installer.py
+++ b/installer.py
@@ -12,9 +12,7 @@ from msg import logos_acknowledge_question
 from msg import logos_error
 from msg import logos_warn
 from utils import check_libs
-from utils import checkDependencies
-from utils import checkDependenciesLogos9
-from utils import checkDependenciesLogos10
+from utils import check_dependencies
 from utils import clean_all
 from utils import cli_download
 from utils import curses_menu
@@ -121,22 +119,19 @@ def chooseVersion():
     QUESTION_TEXT = f"Which version of {config.FLPRODUCT} should the script install?"
     if config.TARGETVERSION is None or config.TARGETVERSION == "":
         options = ["10", "9", "Exit"]
-        versionChoice = curses_menu(options, TITLE, QUESTION_TEXT)
-    else:
-        versionChoice = config.TARGETVERSION
+        version_choice = curses_menu(options, TITLE, QUESTION_TEXT)
     logging.info(f"Target version: {config.TARGETVERSION}")
 
-    checkDependencies()
-    if "10" in versionChoice:
-        checkDependenciesLogos10()
+    if "10" in version_choice:
         config.TARGETVERSION = "10"
-    elif "9" in versionChoice:
-        checkDependenciesLogos9()
+    elif "9" in version_choice:
         config.TARGETVERSION = "9"
-    elif versionChoice == "Exit.":
+    elif version_choice == "Exit.":
         sys.exit(0)
     else:
         logos_error("Unknown version. Installation canceled!", "")
+    
+    check_dependencies()
 
 def logos_setup():
     if config.LOGOS64_URL is None or config.LOGOS64_URL == "":
--- a/utils.py
+++ b/utils.py
@@ -1,5 +1,6 @@
 import atexit
 import curses
+import distro
 import hashlib
 import json
 import logging
@@ -150,7 +151,7 @@ def die_if_running():
     if os.path.isfile(PIDF):
         with open(PIDF, 'r') as f:
             pid = f.read().strip()
-            if cli_continue_question(f"The script is already running on PID {pid}. Should it be killed to allow this instance to run?", "The script is already running. Exiting.", "1"):
+            if cli_continue_question(f"The script is already running on PID {pid}. Should it be killed to allow this instance to run?", "The script is already running. Exiting.", "1", ""):
                 os.kill(int(pid), signal.SIGKILL)
     
     def remove_pid_file():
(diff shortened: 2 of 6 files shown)


thw26:
This will need rebased before review

utils.py
```diff
@@ -481,34 +506,39 @@ def make_skel(app_image_filename):
...
+        subprocess.run([config.SUPERUSER_COMMAND, "steamos-readonly", "disable"], check=True)
+        subprocess.run([config.SUPERUSER_COMMAND, "pacman-key", "--init"], check=True)
+        subprocess.run([config.SUPERUSER_COMMAND, "pacman-key", "--populate", "archlinux"], check=True)
+
+def steam_postinstall_dependencies():
+    subprocess.run([config.SUPERUSER_COMMAND, "sed", '-i', 's/mymachines resolve/mymachines mdns_minimal [NOTFOUND=return] resolve/', '/etc/nsswitch.conf'], check=True)
+    subprocess.run([config.SUPERUSER_COMMAND, "locale-gen"], check=True)
+    subprocess.run([config.SUPERUSER_COMMAND, "systemctl", "enable", "--now", "avahi-daemon"], check=True)
+    subprocess.run([config.SUPERUSER_COMMAND, "systemctl", "enable", "--now", "cups"], check=True)
+    subprocess.run([config.SUPERUSER_COMMAND, "steamos-readonly", "enable"], check=True)
```
thw26, line 519: We could make this more generic:

Have a preinstall() and postinstall(), then check the OS before running the associated commands.

utils.py
```diff
@@ -528,26 +558,23 @@ def check_libs(libraries):
...
+                logos_error(f"The script could not determine your {config.OS_NAME} install's package manager or it is unsupported. Your computer is missing the library: {library}. Please install the package associated with {library} for {config.OS_NAME}.")
+
+def check_dependencies():
+    if int(config.TARGETVERSION) == 10:
+        logging.info("Checking Logos 10 dependencies")
+        install_dependencies(config.PACKAGES)
+    elif int(config.TARGETVERSION) == 9:
+        logging.info("Checking Logos 9 dependencies")
+        install_dependencies(config.PACKAGES)
+        install_dependencies(config.L9PACKAGES)
```
thw26, line 575: I know of at least one person who has continued using Logos 9, and another who keeps trying to go to Logos 10 but reverts to Logos 9.

I would currently consider the Logos 9 code broken in this rewrite. We would need to test this more robustly. We may also need to reconsider how we determine what packages are needed so as to run an install of dependencies only once.

utils.py
```diff
@@ -265,20 +268,30 @@ def get_package_manager():
 
     # Check for package manager and associated packages
     if shutil.which('apt') is not None: # debian, ubuntu
-        config.PACKAGE_MANAGER_COMMAND = "apt install -y"
+        config.PACKAGE_MANAGER_COMMAND_INSTALL = "apt install -y"
+        config.PACKAGE_MANAGER_COMMAND_QUERY = "dpkg -l | grep -E ^"
```
thw26, line 272: Needs testing

utils.py
```diff
@@ -265,20 +268,30 @@ def get_package_manager():
...
         config.PACKAGES = "binutils cabextract fuse wget winbind"
+        config.L9PACKAGES= "" # FIXME: Missing Logos 9 Packages
     elif shutil.which('dnf') is not None: # rhel, fedora
-        config.PACKAGE_MANAGER_COMMAND = "dnf install -y"
+        config.PACKAGE_MANAGER_COMMAND_INSTALL = "dnf install -y"
+        config.PACKAGE_MANAGER_COMMAND_QUERY = "dnf list installed | grep -E ^"
```
thw26, line 277: Needs testing

utils.py
```diff
@@ -265,20 +268,30 @@ def get_package_manager():
...
         config.PACKAGES = "patch mod_auth_ntlm_winbind samba-winbind samba-winbind-clients cabextract bc libxml2 curl"
+        config.L9PACKAGES = "" # FIXME: Missing Logos 9 Packages
     elif shutil.which('yum') is not None: # rhel, fedora
-        config.PACKAGE_MANAGER_COMMAND = "yum install -y"
+        config.PACKAGE_MANAGER_COMMAND_INSTALL = "yum install -y"
+        config.PACKAGE_MANAGER_COMMAND_QUERY = "yum list installed | grep -E ^"
```
thw26, line 282: Needs testing

utils.py
```diff
@@ -265,20 +268,30 @@ def get_package_manager():
...
         config.PACKAGES = "patch mod_auth_ntlm_winbind samba-winbind cabextract bc libxml2 curl"
+        config.L9PACKAGES = "" # FIXME: Missing Logos 9 Packages
     elif shutil.which('pamac') is not None: # manjaro
-        config.PACKAGE_MANAGER_COMMAND = "pamac install --no-upgrade --no-confirm"
+        config.PACKAGE_MANAGER_COMMAND_INSTALL = "pamac install --no-upgrade --no-confirm"
+        config.PACKAGE_MANAGER_COMMAND_QUERY = "pamac list -i | grep -E ^"
```
thw26, line 287: Needs testing.

utils.py
```diff
@@ -208,52 +209,54 @@ def getDialog():
...
+    #    pass
+
+    ## Try reading /etc/debian_version
+    #try:
+    #    with open('/etc/debian_version', 'r') as f:
+    #        config.OS_NAME = 'Debian'
+    #        config.OS_RELEASE = f.read().strip()
+    #        return config.OS_NAME, config.OS_RELEASE
+    #except FileNotFoundError:
+    #    pass
```
thw26, line 253: Once we know that hte below `distro.*()` functions work, we can remove this code.

installer.py
```diff
@@ -121,22 +119,21 @@ def chooseVersion():
     QUESTION_TEXT = f"Which version of {config.FLPRODUCT} should the script install?"
     if config.TARGETVERSION is None or config.TARGETVERSION == "":
         options = ["10", "9", "Exit"]
-        versionChoice = curses_menu(options, TITLE, QUESTION_TEXT)
+        config.TARGETVERSION = curses_menu(options, TITLE, QUESTION_TEXT)
     else:
         versionChoice = config.TARGETVERSION
     logging.info(f"Target version: {config.TARGETVERSION}")
```
n8marti, line 125: It looks like the `else` statement can go away here, if there's no need for the intermediate variable `versionChoice`.

utils.py
n8marti: The `dpkg` command would need to be more like `dpkg -l | grep '^.i  '` (two spaces after the "i"). Or, if we want to go pythonic, we could use the [Python APT library](https://apt-team.pages.debian.net/python-apt/library/index.html) like this, if we're checking if `pkg_name` is installed:
```python
try:
    import apt_pkg
except ModuleNotFoundError: # not debian/ubuntu
    pass
cache = apt_pkg.Cache()
installed_ver = cache[pkg_name].current_ver # None if not installed
```

utils.py
```diff
@@ -287,9 +300,21 @@ def get_runmode():
...
-    command = [config.SUPERUSER_COMMAND, config.PACKAGE_MANAGER_COMMAND] + list(packages)
-    subprocess.run(command, check=True)
+def query_packages(packages):
+    missing_packages = [ ]
+    
+    for p in packages:
+        command = f"{config.SUPERUSER_COMMAND} {config.PACKAGE_MANAGER_COMMAND_QUERY}{p}"
+        result = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True, text=True)
+        if result.returncode != 0:
+            missing_packages.append(p)
```
n8marti, line 310: I don't know about other systems, but ubuntu doesn't need `sudo` to query package status. I think this would require an unnecessary password prompt if the user has all the dependencies already installed.

utils.py
```diff
@@ -287,9 +300,21 @@ def get_runmode():
...
+
+    return missing_packages
+
+def install_packages(packages):
+    command = f"{config.SUPERUSER_COMMAND} {config.PACKAGE_MANAGER_COMMAND_INSTALL} {' '.join(packages)}"
+    print("COMMAND: ", command)
+    subprocess.run(command, shell=True, check=True)
```
n8marti, line 317: maybe you used `print(` here during testing and intended to remove it, but if you want to keep it you should switch it to `logging.info(` to log the command.

utils.py
n8marti: This looks fine for now, unless we find out that other systems have similar kinds of pre/post-install tasks.

utils.py
```diff
@@ -481,34 +506,39 @@ def make_skel(app_image_filename):
...
+def steam_postinstall_dependencies():
+    subprocess.run([config.SUPERUSER_COMMAND, "sed", '-i', 's/mymachines resolve/mymachines mdns_minimal [NOTFOUND=return] resolve/', '/etc/nsswitch.conf'], check=True)
+    subprocess.run([config.SUPERUSER_COMMAND, "locale-gen"], check=True)
+    subprocess.run([config.SUPERUSER_COMMAND, "systemctl", "enable", "--now", "avahi-daemon"], check=True)
+    subprocess.run([config.SUPERUSER_COMMAND, "systemctl", "enable", "--now", "cups"], check=True)
+    subprocess.run([config.SUPERUSER_COMMAND, "steamos-readonly", "enable"], check=True)
+
+def install_dependencies(packages):
+    missing_packages = [ ]
+    package_list = [ ]
```
n8marti, line 523: defining `package_list` as an empty list is redundant if it's defined next as `packages.split()`.

utils.py
```diff
@@ -481,34 +506,39 @@ def make_skel(app_image_filename):
...
+    package_list = packages.split()
+
+    if config.PACKAGE_MANAGER_COMMAND_QUERY:
+        missing_packages = query_packages(package_list)
+        print("MISSING PACKAGES: ", missing_packages )
```
n8marti, line 528: remove `print(` line if no longer needed, or change it to `logging.info(` (or maybe `logging.warning(` in this case?)

utils.py
thw26: The same is true for Arch, and I would imagine the others.

installer.py
thw26: Looks like the config setters below these lines can also go.

thw26: Actually, there is a possibly of the user "exiting" here, so I probably need to do the reverse and have the curses menu set versionChoice.

utils.py
thw26: I'm going to stick with this method since it works for the other distros, but I've made a note in the code to your comment for future consideration.

PR #36, opened 2024-01-05
Allow user to set AppImage Selector.

### LINKED ISSUE #5: AppImage handling
In order to get to beta, we need to make sure our AppImage functionality is properly working as this is how we provide the most stable base install.

We want the new application to be able to download the latest AppImage from the repo, set the AppImage to a newly downloaded exe, or update the appimage in repo to the latest in our repository.

- [x] Convert checkAppImages(). See https://github.com/ferion11/LogosLinuxInstaller/pull/193/commits/bfefb3c05c7a9989e81372d77fa785fc75bd4e94
  - Another way to describe this would be: scan for additional AppImages (in Downloads only?) and add them to the binary list
- [x] Modify AppImage code to check branch version (see #32 and #38)
- [x] Fix set_appimage()
  - [x] This option should only be presented if the user installed using an AppImage. See #56.
  - [x] Update/Modify the config. See https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/36.
  - [x] We should present our recommended AppImages from the FaithLife-Community repo. See https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/5#issuecomment-1878014893. See #54.
  - [x] We should also present those downloaded by the user and which are available in their Logos APPDIR_BINDIR. See https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/36.
- [ ] Add update_appimage()
  - This is similar to `set_appimage()` but bypasses the selection of any binary and simply updates to the latest installed if needed, sort of a quick action set_appimage().
  - [x] This option should only be presented if the user installed using an AppImage. See #56.
  - [ ] The user should be able to click a button or run a command to update to the latest recommended AppImage. See https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/5#issuecomment-1909319121.
  - [x] This also will need the ability to check GitHub for new releases. See https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/5#issuecomment-1878014893. See #54.
  - [x] Make app use the https://github.com/FaithLife-Community/wine-appimages repo. See https://github.com/FaithLife-Community/LogosLinuxInstaller/commit/1d7024a1a93ec07227b4952d1fda53dc15021b3f.
    - [x] Delete AppImage releases from this repo, but only after the above is done
- [x] Detect if AppImage Launcher is running, and, if so, prompt to disable it and warn that a reboot is necessary to proceed. See https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/5#issuecomment-1878009808. See https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/52.

We may want to have Logos launch and automatically check to see if there is a newer (recommended, as in, in repo) AppImage available.

### LINKED ISSUE #32: Cannot Detect If On Devel Branch
@n8marti and I discussed this issue and we do not see a good solution. Beginning with 8.16+ (IIRC) we no longer need to use staging but can use devel. Unfortunately, our method for detecting the branch used can distinguish between staging and stable but not between devel and stable, or put another way, our current check can determine if we are using staging or if stable/devel.

What this means is that the script errors on a valid wine devel version, not presenting it as an option for install. Meaning we are effectively back to using staging.

At present, the only option would be to detect the branch based on an operating system specific check, which we have available with the `get_os()` implemented in the rewrite, but this feels like overkill.

thw26:
This begins the work for #5.

- Allow `set_appimage_symlink` to be reused.
- Create `tui.py` to separate curses functions from utils.py.
- Update `getWineBinOptions()` and `createWineBinaryList()` (renamed).
- Create `find_appimage_files()`.
- Make AppImage selection dynamic.
- Utilize wine branch and version checking.
- Minor clean up, like renaming functions, vars.

~~This PR currently disables the ACTION codes, pending our other discussion, in order to make my testing easier.~~ Fixed in rebase.

#5 will require additional work. We need to have a way of grabbing the latest AppImage release on the FaithLife-Community repo from the program and we need to have a button or means to allow the user to check if they are using the latest recommended AppImage, and if not, to upgrade it.

We may want to move the AppImage releases to a separate repository at FaithLife-Community/Wine-AppImage, which would give us all kinds of new options, such as copying https://github.com/mmtrt/WINE_AppImage repository for our own purposes.

- [x] We need to tackle #32 before this will be fully functional, but we can bring this code in before #32 is fixed. I have verified that the script is finding the appimage files and the binaries, though the binaries fail to report on my system because they are devel rather than staging or stable.
- [x] An outstanding TODO here is to implement Wine AppImage version checks. Since the AppImages produced at https://github.com/mmtrt/WINE_AppImage have the branch name in them, that would make version checking fairly easy. The problem with such an approach will be if a user renames the file.

---

This PR now also partially handles #42 by placing curses in a wrapper.


n8marti:
PR #38 fixes #32 by adding a wine branch check by inspecting `mscoree.dll` in both packaged wine and AppImage wine.

thw26:
Rebased on #40.

thw26:
Current state of this commit is only a partial fix.

Going with our narrower PRs: we can do one for checking branch version, one for modifying the config, one for presenting recommended AppImages, and one for updating the AppImage to the latest GitHub wine appimage. These additional narrower PRs should effectively close #5.

n8marti:
The old name `utils.set_appimage()` function is still referenced in a couple of places: LogosLinuxInstaller.py, L202 (also earlier in the options definition, but maybe that's not a problem); tui_app.py, L69

n8marti:
There is not currently an option for "Set AppImage" in the GUI. Is the wine binary dropdown in the "Install" window sufficient for this?

thw26:
>There is not currently an option for "Set AppImage" in the GUI. Is the wine binary dropdown in the "Install" window sufficient for this?

I think I looked at the control panel and saw you had a button for the function and so didn't add one. I'll look this over in a few hours.

thw26:
I have fixed your review comments and also had to do a lot of code testing in process. (Yay!)

This should be in MUCH better shape.

gui_app.py
```diff
@@ -496,6 +495,33 @@ def run_installer(self, evt=None):
...
 
+    def set_file_dialog_environment(self):
+        desktop_environment = os.environ.get('XDG_CURRENT_DESKTOP', '').lower()
+    
+        if 'gnome' in desktop_environment:
+            os.environ['TK_SILENCE_DEPRECATION'] = '1'
+            os.environ['GTK_FILE_CHOOSER_ACTION'] = 'open'
+        elif 'kde' in desktop_environment:
+            os.environ['TK_SILENCE_DEPRECATION'] = '1'
+            os.environ['KDE_FULL_SESSION'] = 'true'
```
thw26, line 506: This code is not working. The function is called at line 521 but I still see the tk dialog.

n8marti: The function `fd.askopenfilename()` is a Tk function that opens a Tk dialog, so it has no connection with the system's file chooser. I agree that Tk's filechooser is weird and ugly, but I think we'd need to change to a different toolkit if we want a better filechooser. It's possible, but it would bring it more dependencies in addition to having to rework all the GUI code.

One improvement I found, though, is a hack way to hide hidden folders and files:
```python
try:
    # Dummy option below means the function throws an exception, which keeps the window from opening
    # but allows the variables to be set afterwards.
    self.root.tk.call('tk_getOpenFile', '-dummyoption')
except TclError:
        pass
self.root.tk.call('set', '::tk::dialog::file::showHiddenBtn', '1')
self.root.tk.call('set', '::tk::dialog::file::showHiddenVar', '0')

# Then call the fd.askopenfilename() function...
```

gui_app.py
```diff
@@ -496,6 +495,33 @@ def run_installer(self, evt=None):
...
+    def open_file_dialog(self, filetype_name, filetype_extension):
+        root = Tk()
+        root.withdraw()
+    
+        file_path = fd.askopenfilename(
+            title=f"Select {filetype_name}",
+            filetypes=[(filetype_name, f"*.{filetype_extension}"), ("All Files", "*.*")],
+        )
+
+        return file_path
```
n8marti, line 517: I don't think the `root = Tk()` and `root.withdraw()` lines are useful here, if we're content with using `fd.askopenfilename()`.

A hack that I found to hide hidden folders and files:
```python
try:
    # Dummy option below means the function throws an exception, which keeps the window from
    # opening but allows the variables to be set afterwards.
    self.root.tk.call('tk_getOpenFile', '-dummyoption')
except TclError:
    pass
self.root.tk.call('set', '::tk::dialog::file::showHiddenBtn', '1')
self.root.tk.call('set', '::tk::dialog::file::showHiddenVar', '0')

# Then call the fd.askopenfilename() function...
```
It looks like that code could be worked into this `open_file_dialog` function, and then the function could become a module-level function outside of the `ControlWindow` class. That would allow it to be reused in other cases. It might also be useful to have an analogous `open_folder_dialog` function to go with it.

gui_app.py
```diff
@@ -496,6 +495,33 @@ def run_installer(self, evt=None):
...
+
+    def set_appimage(self, evt=None):
+        Tk().withdraw()
```
n8marti, line 520: I think `Tk().withdraw()` needs to be removed here.

gui_app.py
```diff
@@ -496,6 +495,33 @@ def run_installer(self, evt=None):
...
+        self.set_file_dialog_environment()
+        appimage_filename = self.open_file_dialog("AppImage", "AppImage")
```
n8marti, line 522: If the user cancels the dialog, it returns an empty tuple `()`. Maybe the best way to handle that is to add something like:
```python
if not appimage_filename:
    return
```

installer.py
```diff
@@ -152,9 +152,10 @@ def chooseInstallMethod():
 
     if config.WINE_EXE is None:
         logging.info("Creating binary list.")
+        appimages = find_appimage_files()
```
n8marti, line 155: maybe add a line here to for `logging.debug` to list the found appimages?

utils.py
```diff
@@ -493,8 +552,80 @@ def cli_download(uri, destination):
...
+    directories = [config.APPDIR_BINDIR, get_user_downloads_dir(), directory]
+    for dir in directories:
+        for root, dirs, files in os.walk(dir):
+            for filename in files:
+                if filename.lower().startswith("wine") and filename.lower().endswith("appimage"):
+                    appimage_path = os.path.join(root, filename)
+                    if os.access(appimage_path, os.X_OK):
+                        appimages.append(appimage_path)
+                    else:
+                        pass
```
n8marti, line 626: `pathlib.Path()` is very compact and efficient when it comes to recursive searching. You could do something like:
```python
        appimage_paths = Path(dir).rglob('wine*.appimage', case_sensitive=False)
        for p in appimage_paths:
            if os.access(p, os.X_OK):
                appimages.append(p)
```

utils.py
```diff
@@ -597,6 +728,21 @@ def file_exists(file_path):
...
+
+def check_file_type(file_path, file_type):
+    if not file_path.endswith(f".{file_type}"):
+        raise argparse.ArgumentTypeError(f"Invalid file type: '{file_path}'. Must be of type '{file_type}.'")
+    return file_path
+
+def check_appimage(file_path):
+    file_path = check_file_exists(file_path)
+    file_path = check_file_type(file_path, "AppImage")
+    return file_path
```
n8marti, line 744: I think `check_appimage()` could use `file_exists()` also defined here in `utils.py` rather than adding `check_file_exists()`. It would need to handle a `False` result.

In general, we've tended to handle incompatible variable values by returning `False` rather than raising an exception, both of which need to then be handled anyway. Any reason not to just return `False` in `check_file_type()`?

thw26:
This is not ready. It requires updates to `installer.beginInstall()`.

utils.py
```diff
@@ -285,31 +302,31 @@ def get_package_manager():
...
+        config.L9PACKAGES = ""  # FIXME: Missing Logos 9 Packages
+    elif shutil.which('pamac') is not None:  # manjaro
         config.PACKAGE_MANAGER_COMMAND_INSTALL = "pamac install --no-upgrade --no-confirm"
         config.PACKAGE_MANAGER_COMMAND_QUERY = "pamac list -i | grep -E ^"
         config.PACKAGES = "patch wget sed grep gawk cabextract samba bc libxml2 curl"
-        config.L9PACKAGES = "" # FIXME: Missing Logos 9 Packages
-    elif shutil.which('pacman') is not None: # arch, steamOS
+        config.L9PACKAGES = ""  # FIXME: Missing Logos 9 Packages
+    elif shutil.which('pacman') is not None:  # arch, steamOS
         config.PACKAGE_MANAGER_COMMAND_INSTALL = 'pacman -Syu --overwrite \* --noconfirm --needed'
```
n8marti, line 326: python3.12 gives a `SyntaxWarning` about `\*` in this line. I making it a raw string; i.e. `r'pacman ...'` . I did this already for the reg file strings that also have backslashes in them.

tui.py
```diff
@@ -0,0 +1,165 @@
...
+    screen = curses.initscr()
+    curses.echo()
+    screen.addstr(0, 0, question_text)
+    screen.refresh()
+    user_input = screen.getstr(1, 0, 50).decode('utf-8')
+    curses.noecho()
+    return user_input
+
+
+def confirm(title, question_text):
```
n8marti, line 16: I got this confirmation dialog after choosing the AppImage in my Downloads folder. I hit [Enter] to see if 'Y' was the default answer, which is usually what's implied when it's capitalized `[Y/n]`, but then I got the error `Error running TUI control panel: [Errno 2] No such file or directory: 'wine-devel_8.19-x86_64.AppImage'`.

Because of the error, the script exited with the terminal wordwrap being screwed up and the cursor being invisible. So can you add in the curses wrapper function here?

But the original error needs to be handled, too. Is it a matter of the file path only being the name and not the full path?

config.py
```diff
@@ -81,13 +82,14 @@
...
+L9PACKAGES = None
 SUPERUSER_COMMAND = None
 
 persistent_config_keys = [
     "FLPRODUCT", "FLPRODUCTi", "TARGETVERSION", "INSTALLDIR", "APPDIR",
     "APPDIR_BINDIR", "WINETRICKSBIN", "WINEPREFIX", "WINEBIN_CODE", "WINE_EXE",
-    "WINESERVER_EXE", "WINE64_APPIMAGE_FULL_URL", "WINECMD_ENCODING",
-    "WINE64_APPIMAGE_FULL_FILENAME", "APPIMAGE_LINK_SELECTION_NAME",
+    "WINESERVER_EXE", "APPIMAGE_FILENAME", "RECOMMENDED_WINE64_APPIMAGE_FULL_URL", "WINECMD_ENCODING",
+    "RECOMMENDED_WINE64_APPIMAGE_FULL_FILENAME", "APPIMAGE_LINK_SELECTION_NAME",
```
n8marti, line 92: Do we need to save `RECOMMENDED_WINE64_APPIMAGE` variables to the config file? It seems like they would be fine just being script/app global variables, right? It would also avoid any potential problems down the road of the recommended values change but the user's config keeps the old values. And I'm in favor of reducing the config file to only the fewest necessary variables.

LogosLinuxInstaller.py
```diff
@@ -181,13 +180,13 @@ def parse_args(args, parser):
...
+        config.ACTION = wine.run_logos()
     elif args.run_indexing:
-        wine.run_indexing
+        wine.run_indexing()
     elif args.remove_library_catalog:
-        control.remove_library_catalog
+        control.remove_library_catalog()
     elif args.remove_index_files:
-        control.remove_all_index_files
+        control.remove_all_index_files()
```
n8marti, line 189: The functions should be used to set `config.ACTION` here but left uncalled because the rest of the runtime config is not yet setup. The defined action is then called later in `main`.

LogosLinuxInstaller.py
```diff
@@ -197,7 +196,8 @@ def parse_args(args, parser):
     elif args.restore:
         config.ACTION = control.restore
     elif args.set_appimage:
-        utils.set_appimage
+        appimage_file_path = args.set_appimage[0]
+        utils.set_appimage_symlink(appimage_file_path, True)
```
n8marti, line 200: Following on from the above comment, since the action function should only be assigned and not yet called, it needs to take zero parameters. So the definition of `utils.set_appimage_symlink` should take zero parameters, or there should be a "wrapper" function taking zero parameters that then includes L199-200 to properly set up and call `utils.set_appimage_symlink`.

gui_app.py
```diff
@@ -12,8 +12,7 @@
 
 from tkinter import Tk
 from tkinter import Toplevel
-from tkinter.filedialog import askdirectory
-from tkinter.filedialog import askopenfilename
+from tkinter import filedialog as fd
```
n8marti, line 15: L528's `askdirectory` function is orphaned with this change. I guess it just needs to be changed to `fd.askdirectory`.

utils.py
```diff
@@ -496,7 +427,7 @@ def cli_download(uri, destination):
     # Download from uri in thread while showing progress bar.
     cli_queue = queue.Queue()
     args = [uri, target]
-    kwargs = kwargs={'q': cli_queue}
+    kwargs = kwargs = {'q': cli_queue}
```
n8marti, line 430: Not sure how we got `kwargs` defined twice, but this just needs to be ` kwargs = {'q': cli_queue}`

utils.py
```diff
@@ -863,21 +856,16 @@ def same_size(url, file_path):
...
-def is_appimage(file_path):
-    # Ref:
-    # - https://cgit.freedesktop.org/xdg/shared-mime-info/commit/?id=c643cab25b8a4ea17e73eae5bc318c840f0e3d4b
-    # - https://github.com/AppImage/AppImageSpec/blob/master/draft.md#image-format
-    # Note: 
-    # result is a tuple: (is AppImage: True|False, AppImage type: 1|2|None)
-    result = (False, None)
     file_path = Path(file_path).expanduser().resolve()
     with file_path.open('rb') as f:
         f.seek(1)
```
n8marti, line 871: these lines got orphaned when `is_appimage` was moved; or, rather, they got grafted into `write_progress_bar`.

utils.py
```diff
@@ -935,3 +923,177 @@ def get_latest_folder(folder_path):
...
+    directories = [
+        os.path.expanduser("~") + "/bin",
+        config.APPDIR_BINDIR,
+        get_user_downloads_dir()
+    ]
+    if config.CUSTOMBINPATH is not None:
+        directories.append(config.CUSTOMBINPATH)
+
+    for dir in directories:
+        appimage_paths = find_paths(dir, r'wine.*\.appimage')
```
n8marti, line 1002: Is this the only place `find_paths` is used? If so, then a simple `Path().rglob('wine*.appimage', case_sensitive=False)` is sufficient, and should be significantly faster than `find_paths`, which checks *every* file and subfolder in `directory` against `regex_expression`.

n8marti (commented):
This covers a lot of ground in organizing things a bit better. Good work on that. I'm curious if you started using some kind of linter? Because there are tons of changes that are simply additional line breaks between functions and single spaces added before comment '#'s.

Once we've resolved the comments I'll do some testing before being comfortable with merging it in.

config.py
thw26: Yes—let's do that.. I'll also remove APPIMAGE_LINK_SELECTION_NAME, which can be a system level var.

LogosLinuxInstaller.py
thw26: Ah, okay. I got an error when running these and the solution was to add the parentheses. I will change all of these around.

utils.py
thw26: Python kept complaining about this for some reason. I will change it back.

thw26:
> This covers a lot of ground in organizing things a bit better. Good work on that. I'm curious if you started using some kind of linter? Because there are tons of changes that are simply additional line breaks between functions and single spaces added before comment '#'s.
> 
> Once we've resolved the comments I'll do some testing before being comfortable with merging it in.

I installed PyCharm for an IDE as things started getting a little unwieldy for `vim`; I didn't want to have to keep track of variable names and I was wanting to change several. PyCharm is making use of PEP 8, and so it alerts to errors, makes lots of warnings about issues, and then also makes many soft warnings for things like PEP 8 violations.

The last thing I need to do in your comments is the curse wrapper.

tui.py
thw26: I've added wrapper lines to each function.

Capitalizing just one also implies case insensitive. On some bash prompts, `yes` will be defaulted, but usually the default will be given in parentheses `(default=Y)`. Here I use `[Y/n]` to mean case insensitive.

Now, as regards the file name, I'm guessing you mean you are testing `utils.set_appimage_symlink()`. Line 1051 (only after I push the changes?) should be the only line that would call the file name:

```
appimage_filepath = f"{config.APPDIR_BINDIR}/{appimage_filename}" 
```

That would suggest that `config.APPDIR_BINDIR` isn't being added to the file path.

thw26: I did at one time have a function for requiring an absolute path, but I didn't think it was necessary because the function has `check_appimage` which determines whether the appimage exists at all.

thw26: Although, now that I think about it, the issue may lie here:

```
    if confirm is True or confirm == 'yes':
        logging.info(f"Copying AppImage to {config.APPDIR_BINDIR}.")
        shutil.copy(selected_appimage, f"{config.APPDIR_BINDIR}")
        os.symlink(appimage_filename, appimage_symlink_path)
        config.SELECTED_APPIMAGE_FILENAME = appimage_filename
```

`appimage_filename` in the symlink command should probably be `{config.APPDIR_BINDIR}/{appimage_filename}`. I think I'm wrongly assuming that I have switched my working directory there to the `config.APPDIR_BINDIR`.

thw26: In committed changes right now, I did make one modification to the default var as part of this, which uses config.RECOMMENDED_WINE64_APPIMAGE_FULL_FILENAME instead of config.RECOMMENDED_WINE64_APPIMAGE_FILENAME, the only difference being whether it has the extension or not.

thw26: I just pushed some unsquashed changes in a third commit attempting to resolve this issue.

PR #40, opened 2024-01-08
Rework startup logic

### LINKED ISSUE #37: Separate Optargs
Per #34, we need to separate optargs as being either runtime variable setters or actions/commands.

The Bash script did this by iterating over the optargs twice, once for setters, once for actions.

How should we best handle this?

Should the GUI app be made into one panel rather than two, by making "install" an option on the control panel?

n8marti:
Fix #37.

* Clean up handling of runtime options vs installed app subcommands.
* Clean up `--help` output.
* Change GUI color scheme to better match Logos.
* Move TUI code to its own module.


LogosLinuxInstaller.py
```diff
@@ -5,74 +5,156 @@
...
+        '--make-skel', action='store_true',
+        #help='make a skeleton install only',
+        help=argparse.SUPPRESS,
+    )
+    cmd.add_argument(
+        '--check-resources', action='store_true',
+        #help='check resources'
+        help=argparse.SUPPRESS,
+    )
+    return parser
```
thw26, line 143: I noticed in the other commits that you reduced the style for the GUI app to single lines rather than multilines.

While I tend to prefer single lines, is there a reason these are kept as multiline?

LogosLinuxInstaller.py
```diff
@@ -102,153 +175,124 @@ def parse_args(args):
...
+    elif args.run_winetricks:
+        config.ACTION = wine.run_winetricks
+    elif args.toggle_app_logging:
+        config.ACTION = wine.switch_logging
+    elif args.create_shortcuts:
+        config.ACTION = installer.create_shortcuts
+    elif args.remove_install_dir:
+        config.ACTION = control.remove_install_dir
+    else: # default function if app is installed
+        config.ACTION = run_control_panel
```
thw26, line 215: Do we need to do checks that verify these things are installed? As in, error if no Logos install found?

wine.py
```diff
@@ -30,24 +27,13 @@ def get_pids_using_file(file_path, mode=None):
...
-    logging.info(f"* Starting wait_process_using_dir for {directory}…")
-
-    # Get pids and wait for them to finish.
-    pids = get_pids_using_file(directory)
-    for pid in pids:    
-        logging.info(f"wait_process_using_dir PID: {pid}")
-        psutil.wait(pid)
-
-    logging.info("* End of wait_process_using_dir.")
-
```
thw26, line 43: Is there a reason this is removed?

Just unused?

thw26 (commented):
Most of this is good to go. I like the color changes to the GUI; a nice touch.

LogosLinuxInstaller.py
n8marti: I was having a hard time scanning back and forth, trying to confirm all of the parameters. Making in multiline made that easier. It also allowed adding comments mid-function. Plus, I tend to like to break up long lines, which we have with the inclusion of the "help" parameters. I'm not married to this formatting, though.

wine.py
n8marti: Ha, for some reason the same function was both imported from `utils.py` to and then defined in `wine.py`. So this definition was redundant.

PR #50, opened 2024-01-18
Add `-K|--skip-dependencies`. Fix #49.

### LINKED ISSUE #49: Add a `--skip-dependencies` Flag
This would be very helpful in testing and would serve a similar purpose as `--skip-fonts`.

thw26:
Fix #49.

### DIFF
--- a/LogosLinuxInstaller.py
+++ b/LogosLinuxInstaller.py
@@ -28,6 +28,10 @@ def get_parser():
         help='skip font installations',
     )
     cfg.add_argument(
+        '-K', '--skip-dependencies', action='store_true',
+        help='skip dependencies check and installation',
+    )
+    cfg.add_argument(
         '-V', '--verbose', action='store_true',
         help='enable verbose mode',
     )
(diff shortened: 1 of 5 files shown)


n8marti:
I think the code looks good, and I tested it as far as seeing that the GUI checkbox looks to be working correctly. But I'm wondering if having it in the GUI is unnecessary? This isn't something that will be used in 95% of cases, right? I would vote for it to be a CLI-only option, but since the box *is* auto-checked unless the option is passed, I'm also okay with it staying in the GUI. I'll leave the final decision up to you.

n8marti (approved):
I think the code looks good, and I tested it as far as seeing that the GUI checkbox looks to be working correctly. But I'm wondering if having it in the GUI is unnecessary? This isn't something that will be used in 95% of cases, right? I would vote for it to be a CLI-only option, but since the box is auto-checked unless the option is passed, I'm also okay with it staying in the GUI. I'll leave the final decision up to you.

Issue #19, opened 2023-12-30
Resume Interrupted Install


thw26:
https://github.com/ferion11/LogosLinuxInstaller/issues/162

PR #52, opened 2024-01-25
Add ability to detect and remove AppImageLauncher

thw26:
Partial fix for #5. This adds the ability to remove AppImage Launcher.

Add `config.PMC_REMOVE`; remove yum
Add `config.BADPACKAGES`
Add `query_package()` "mode" var, add `remove_packages()`, add `check_bad_packages()`

I've also made sure to add a major warning label to this remove command to its destructive capability.

One thing we may need to add is an exit function to the AppImageLauncher detection. If it is found, Logos won't be able to launch, and the install should fail at that point, otherwise we will wind up having support requests.

If the user removes it, the install should exit and instruct the user to reboot, or offer to reboot the system for the user but with warning.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -194,6 +194,12 @@ def die(message):
     logging.critical(message)
     sys.exit(1)
 
+def reboot():
+    logging.info("Rebooting system.")
+    command = f"{config.SUPERUSER_COMMAND} reboot now"
+    subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True, text=True)
+    sys.exit(0)
+
 
 def set_verbose():
     config.LOG_LEVEL = logging.INFO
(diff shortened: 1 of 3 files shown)


utils.py
```diff
@@ -345,25 +348,36 @@ def get_runmode():
...
+    bad_packages = []
 
     for p in packages:
         command = f"{config.PACKAGE_MANAGER_COMMAND_QUERY}{p}"
         logging.debug(f"pkg query command: \"{command}\"")
         result = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True, text=True)
         logging.debug(f"pkg query result: {result.returncode}")
-        if result.returncode != 0:
+        if result.returncode != 0 and mode == "install":
             missing_packages.append(p)
```
n8marti, line 364: You can simplify this function by setting a default value for mode:
```python
def query_packages(packages, mode='install'):
```
This would remove the `elif` code on L367-368 and make passing `mode` optional when calling it.

utils.py
```diff
@@ -553,6 +577,28 @@ def install_dependencies(packages):
...
 
+def check_bad_packages(packages):
+        bad_packages = []
+        package_list = packages.split()
+
+        if config.PACKAGE_MANAGER_COMMAND_QUERY:
+            bad_packages = query_packages("remove", package_list)
+
+        if config.PACKAGE_MANAGER_COMMAND_INSTALL:
+            message = f"Your {config.OS_NAME} computer may have bad packages installed. To continue, the script will attempt to remove the package(s): {bad_packages} by using ({config.PACKAGE_MANAGER_COMMAND_INSTALL}).\n\nWARNING: Automating the removal of packages could leave your system in an unusable state if the removal also removes system-required dependencies.\n\nProceed?"
```
n8marti, line 588: Would it be a little clearer to call the packages "conflicting" packages rather than "bad" when communicating with the user?

utils.py
```diff
@@ -345,25 +348,36 @@ def get_runmode():
...
+        msg = 'None'
+        if missing_packages:
+            msg = f"Missing packages: {' '.join(missing_packages)}"
+        logging.info(f"Missing packages: {msg}")
+        return missing_packages
+    if mode == "remove":
+        msg = 'None'
+        if bad_packages:
+            msg = f"Bad packages: {' '.join(bad_packages)}"
+        logging.info(f"Bad packages: {msg}")
```
n8marti, line 380: Minor detail: You can define `msg = 'None'` once just before `if mode == "install":` rather than doing it twice (once after each `if mode ==` statement).

n8marti:
> One thing we may need to add is an exit function to the AppImageLauncher detection. If it is found, Logos won't be able to launch, and the install should fail at that point, otherwise we will wind up having support requests.

I think this would fit logically around the same time as (or somehow incuded in?) when the dependency checks are run.

n8marti:
> If the user removes it, the install should exit and instruct the user to reboot, or offer to reboot the system for the user but with warning.

I don't know how portable this is, but in debian-based systems if you want to require a reboot you can [create the file](https://www.debian.org/doc/debian-policy/ch-opersys.html#signaling-that-a-reboot-is-required) `/run/reboot-required`, which will (eventually?) pop up a window saying "Your system needs to be restarted", and giving the Restart/Now now/Cancel options.

utils.py
thw26: Agreed. I wanted to think of a better term but didn't consider conflict.

thw26:
>I think this would fit logically around the same time as (or somehow incuded in?) when the dependency checks are run.

Yes, that makes the most sense.

At present I don't actually remove the packages because it can be destructive if the system thinks it needs to remove some lower dependency. Should we require the user to remove manually or automate it, since this is designed for automated install?

thw26:
> I don't know how portable this is, but in debian-based systems if you want to require a reboot you can [create the file](https://www.debian.org/doc/debian-policy/ch-opersys.html#signaling-that-a-reboot-is-required) `/run/reboot-required`, which will (eventually?) pop up a window saying "Your system needs to be restarted", and giving the Restart/Not now/Cancel options.

That would be nice to do. The CLI would be easy to do this with. The GUI could probably do it, too. We would just need to determine a user's init system (e.g., systemd or otherwise) and then hardcode the command.

Would something like this play into #19?

n8marti:
> > I think this would fit logically around the same time as (or somehow incuded in?) when the dependency checks are run.
> 
> Yes, that makes the most sense.
> 
> At present I don't actually remove the packages because it can be destructive if the system thinks it needs to remove some lower dependency. Should we require the user to remove manually or automate it, since this is designed for automated install? We could just present the user with a string they need to run (e.g., "Please run XXX in your terminal.", or direct them to their software repos.
> 
> Personally I think we should do automated removal, but with warnings.
> 
> Also, I will need to wrap install_dependencies and the new check_conflicting_packages into one new function, or find a better way to combine them. I'm leaning toward the latter, as it should mean less overhead, especially for any SteamOS users. I could then also minimize user interaction by only having one request to proceed.

I think I'd be fine with either warning the user to uninstall AppImage Launcher before running LogosLinuxInstaller again, or auto-uninstalling it using LogosLinuxInstaller. Surely there aren't other apps that are depending on it, except that I'd want [someone] to test distros like Manjaro that have in the OS ISO. For example, I know that if you install ubuntu server, then later install ubuntu-desktop on top of it, ubuntu-desktop "depends" on all the default desktop apps. So if you later try to uninstall one of the ubuntu-desktop dependencies, then it will also want to uninstall ubuntu-desktop itself, which will have a huge cascading effect. If that unwelcome scenario is possible for AppImage Launcher on distros like Manjaro, I'd be a little worried about removing AppImage Launcher. But again, that doesn't seem *too* likely, right?

thw26:
>But again, that doesn't seem too likely, right?

My sentiments exactly.

n8marti:
> > I don't know how portable this is, but in debian-based systems if you want to require a reboot you can [create the file](https://www.debian.org/doc/debian-policy/ch-opersys.html#signaling-that-a-reboot-is-required) `/run/reboot-required`, which will (eventually?) pop up a window saying "Your system needs to be restarted", and giving the Restart/Not now/Cancel options.
> 
> That would be nice to do. The CLI would be easy to do this with. The GUI could probably do it, too. We would just need to determine a user's init system (e.g., systemd or otherwise) and then hardcode the command.
> 
> Would something like this play into #19?

Hm, to accomplish resumable install, I'd be tempted to make the install steps very explicit in the script, then have each step simply check to see whether whatever it's supposed to do has already been done or not. If so, it skips on to the next one; if not, is runs that step. That would replace the current strategy of just checking for the INSTALLDIR and refusing to continue if that folder exists.

thw26:
> Hm, to accomplish resumable install, I'd be tempted to make the install steps very explicit in the script, then have each step simply check to see whether whatever it's supposed to do has already been done or not. If so, it skips on to the next one; if not, is runs that step. That would replace the current strategy of just checking for the INSTALLDIR and refusing to continue if that folder exists.

That's more or less what I was thinking for how to do this.

I would think this would be necessary if we are going to require a reboot for AIL.

thw26:
Given the above, I'll update the code to combine the two dependency check functions. I will probably stub the reboot functionality and I might just have it exit with a message for the time being as a later PR.

n8marti:
I haven't had time to do any further testing, but I don't have any further code comments after this most recent commit.

Issue #46, opened 2024-01-15
Add A Self-Update Command


thw26:
It would be great if we had our own built-in update utility for our binary. This would make it even easier for the average user.

I would think the work we do on implementing #5 would help us accomplish this same task, namely pulling updates from GitHub.

PR #54, opened 2024-02-01
Dynamically set recommended appimage URL and config

n8marti:
Partial fix for #5:
- set recommended appimage based on latest release from FaithLife-Community repo

Bonus:
- a couple of small fixes for dependency checking


utils.py
```diff
@@ -977,6 +982,22 @@ def get_latest_folder(folder_path):
     return latest
 
 
+def set_recommended_appimage_config():
+    releases_url = "https://api.github.com/repos/FaithLife-Community/wine-appimages/releases"  # noqa: E501
+    json_data = json.loads(net_get(releases_url))
+    appimage_url = json_data[0].get('assets')[0].get('browser_download_url')
+    logging.info(f"Recommended AppImage URL: {appimage_url}")
+    config.RECOMMENDED_WINE64_APPIMAGE_FULL_URL = appimage_url
+    config.RECOMMENDED_WINE64_APPIMAGE_FULL_FILENAME = os.path.basename(appimage_url)  # noqa: E501
```
thw26, line 991: Could we turn this into a generic function for getting release info from GitHub, such that we could apply it to the wine-appimages repo and the installer repo? This way the function could supply code to both getting the recommended appimage for #5 and also for #46?

`set_recommended_appimage_config()` would then call this generic getter function and set the config to the latest release by running the succeeding parsing code.

thw26 (commented):
Could we

utils.py
thw26: My thoughts then are that we could work a second function that also sets the config with the latest installer release, too. We could run this function in the background or during start to check if there is a new release for either.

thw26:
I just got the following error while working in the code:

```
Installer log file: /home/thwright/.local/state/Logos_on_Linux/Logos_on_Linux.log
  File "/home/thwright/Dev/LogosLinuxInstaller/./LogosLinuxInstaller.py", line 313, in <module>
    main()
  File "/home/thwright/Dev/LogosLinuxInstaller/./LogosLinuxInstaller.py", line 244, in main
    utils.set_default_config()
  File "/home/thwright/Dev/LogosLinuxInstaller/utils.py", line 136, in set_default_config
    set_recommended_appimage_config()
  File "/home/thwright/Dev/LogosLinuxInstaller/utils.py", line 988, in set_recommended_appimage_config
    appimage_url = json_data[0].get('assets')[0].get('browser_download_url')
                   ~~~~~~~~~^^^
KeyError: 0
```

thw26:
Here's the issue:

```
2024-02-07 22:24:32 DEBUG: {'message': "API rate limit exceeded for ip.ad.dr.ess. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)", 'documentation_url': 'https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting'}
```

utils.py
```diff
@@ -977,6 +982,35 @@ def get_latest_folder(folder_path):
...
+def get_latest_release_url(releases_url):
+    appimage_url = None
+    json_data = json.loads(net_get(releases_url))
+    logging.debug(f"{json_data=}")
+    if json_data is not None:
+        appimage_url = json_data[0].get('assets')[0].get('browser_download_url')  # noqa: E501
+    else:
+        logging.critical("Could not get latest appimage URL.")
+    logging.info(f"Recommended AppImage URL: {appimage_url}")
+    return appimage_url
```
thw26, line 994: Could you make these var names more generic, since we may not always be seeking an AppImage?

utils.py
```diff
@@ -977,6 +982,35 @@ def get_latest_folder(folder_path):
...
+
+
+def set_recommended_appimage_config():
+    releases_url = "https://api.github.com/repos/FaithLife-Community/wine-appimages/releases"  # noqa: E501
+    appimage_url = get_latest_release_url(releases_url)
+    if appimage_url is None:
+        logging.critical("Unable to set recommended appimage config without URL.")  # noqa: E501
+        return
```
thw26, line 1002: Since this code requires a very specific string format to work, could you simply document what we currently use for the sake of reference?

thw26 (changes requested):

thw26 (approved):

PR #56, opened 2024-02-10
Disable setappimage if winebincode != AppImage

thw26:
This is a partial fix for #5. It adds conditional logic to the CLI, TUI, and GUI that either removes or disables a function if the config file states that the Logos install was created with something other than an AppImage.

It also creates a new GUI tooltip function. This is applied to the set appimage button, so that when the user hovers over it with his mouse, it will tell him why the button is disabled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,7 @@ __pycache__/
 dist/
 build/
 env/
+.env/
 venv/
+.venv/
 .idea/
--- a/LogosLinuxInstaller.py
+++ b/LogosLinuxInstaller.py
@@ -203,11 +203,14 @@ def parse_args(args, parser):
         config.ACTION = control.restore
     elif args.set_appimage:
         config.APPIMAGE_FILE_PATH = args.set_appimage[0]
-        if not utils.file_exists(config.APPIMAGE_FILE_PATH):
-            raise argparse.ArgumentTypeError(f"Invalid file path: '{config.APPIMAGE_FILE_PATH}'. File does not exist.")
-        if not utils.check_appimage(config.APPIMAGE_FILE_PATH):
-            raise argparse.ArgumentTypeError(f"{config.APPIMAGE_FILE_PATH} is not an AppImage.")
-        config.ACTION = utils.set_appimage_symlink
+        if config.WINEBIN_CODE == "AppImage":
+            if not utils.file_exists(config.APPIMAGE_FILE_PATH):
+                raise argparse.ArgumentTypeError(f"Invalid file path: '{config.APPIMAGE_FILE_PATH}'. File does not exist.")
+            if not utils.check_appimage(config.APPIMAGE_FILE_PATH):
+                raise argparse.ArgumentTypeError(f"{config.APPIMAGE_FILE_PATH} is not an AppImage.")
+            config.ACTION = utils.set_appimage_symlink
+        else:
+            msg.logos_error("The command you requested is disabled. The configured install was not created using an AppImage.")
     elif args.get_winetricks:
         config.ACTION = control.get_winetricks
     elif args.run_winetricks:
--- a/gui.py
+++ b/gui.py
@@ -1,3 +1,4 @@
+from tkinter import Toplevel
 from tkinter import BooleanVar
 from tkinter import font
 from tkinter import IntVar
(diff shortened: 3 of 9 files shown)


n8marti:
The general logic looks good.

I just started testing with the Tk dialog and ran into this:
```
Exception in Tkinter callback
  File "/opt/lib/python3.12/tkinter/__init__.py", line 1962, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "/home/nate/g/LogosLinuxInstaller/gui_app.py", line 228, in on_wine_check_released
    self.gui.wine_dropdown['values'] = utils.get_wine_options(utils.find_appimage_files(), utils.find_wine_binary_files())
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nate/g/LogosLinuxInstaller/utils.py", line 757, in get_wine_options
    return sorted(wine_binary_options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'PosixPath' and 'str'
```
It looks like the list `wine_binary_options` is a mix between strings and pathlib.Paths. It looks like `utils.find_appimage_files()` returns pathlib.Paths, while `utils.find_wine_binary_files()` returns strings. Both of those get passed into `utils.get_wine_options()`, so we either need to change what type one of them returns, or we need to covert it in `utils.get_wine_options()` to be the same as the other one.

thw26:
> It looks like the list `wine_binary_options` is a mix between strings and pathlib.Paths. It looks like `utils.find_appimage_files()` returns pathlib.Paths, while `utils.find_wine_binary_files()` returns strings. Both of those get passed into `utils.get_wine_options()`, so we either need to change what type one of them returns, or we need to covert it in `utils.get_wine_options()` to be the same as the other one.

Okay. That should be an easy fix; I'll mod find_appimage_files. If there is a need to have both returned, I'll add a flag for the function(s).

thw26:
I was running my system's default Python version, 3.11, and I hit an error in code. I was about to code around this until I checked docs and saw it was a 3.12 feature added. While I don't plan to do a full-blown version check for every feature, I added this for clarity's sake since the CLI reported an unexpected keyword error. Thus I added this commit: [229bcb8](https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/56/commits/229bcb8d9e8c4b60a01c319600913ced065d63ba).

n8marti:
```
Exception in thread Thread-2 (get_logos_releases):
  File "/opt/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "/opt/lib/python3.12/threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File "/home/nate/g/LogosLinuxInstaller/utils.py", line 702, in get_logos_releases
    filtered_releases = filter_versions(releases, 30, 1)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nate/g/LogosLinuxInstaller/utils.py", line 671, in filter_versions
    return [version for version in versions if check_logos_release_version(version, threshold, check_version_part)]
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nate/g/LogosLinuxInstaller/utils.py", line 667, in check_logos_release_version
    return version_parts[check_version_part - 1] < threshold
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'map' object is not subscriptable
```
So it apparently doesn't work to pluck an item by index from the `map` function object without first converting that into a list, in which case, I'd say just make it a list from the beginning. So L666 could be something like:
```
version_parts = [int(p) for p in version.split('.')]
```
or:
```
version_parts = list(map(int, version.split('.')))
```
I tested the speed of both of these, just out of curiosity. Here's what I found:
```python
>>> import timeit
>>> timeit.timeit('list(map(int, "30.2.001".split(".")))', number=10000)
0.014728516049217433
>>> timeit.timeit('[int(p) for p in "30.2.001".split(".")]', number=10000)
0.011726382013875991
```
So the list comprehension option seems to be faster, in case you ever want to do this on a larger number of items.

thw26:
Ah, I had simplified that code from if/elif series/case statement to that. Sounds good. Will test a change and push.

n8marti (approved):
Testing a clean install for TUI and GUI both succeed. This looks good to me.

PR #62, opened 2024-02-19
Add Control Panel desktop panel icon

### LINKED ISSUE #12: Tk window needs an app icon
When LogosLinuxInstaller is run with the Tk GUI there is a blank icon in the panel. This should be fixable in the LogosLinuxInstaller.spec file.

n8marti:
- include `img/logos4-128-icon.png` in built binary
- set Control Panel icon as the included file; fixes #12

I chose to include the Logos icon and always use it for the Control Panel because it seems strange to not have a Control Panel icon until the user has chosen which Faithlife app to install (Logos/Verbum). I guess another option could be to use a more generic FaithLife icon for the Control Panel, but with the risk of confusing most users? Feedback is welcome.


n8marti:
It now should default to the Logos icon but then update to the Verbum icon if that app is installed instead. I haven't been able to test that, because I don't have any Verbum installers already downloaded.

thw26 (approved):

PR #63, opened 2024-02-19
Add Update To Latest AppImage Option

### LINKED ISSUE #5: AppImage handling

thw26:
This commit resolves #5. It adds the necessary functions to compare the current AppImage version and to update the user's install to the latest AppImage. This button is conditionally visible depending on if WINEBIN_CODE == AppImage.

It also now attempts to circumvent hitting the GitHub API limit by catching the error in `net_get()` and adding a `utils.self_update()` function.

When this code is submitted, we can release alpha3.

This also lays groundwork for #46.


thw26:
This code should now be functional for the CLI and the TUI.

If you can confirm, I will then code the GUI.

LogosLinuxInstaller.py
```diff
@@ -201,6 +205,17 @@ def parse_args(args, parser):
         config.ACTION = control.backup
     elif args.restore:
         config.ACTION = control.restore
+    elif args.update_latest_appimage:
+        if config.WINEBIN_CODE == "AppImage" or config.WINEBIN_CODE == "Recommended":
+            status = utils.compare_recommended_appimage_version()
```
n8marti, line 210: `utils.compare_recommended_appimage_version()` returns a tuple, so maybe it should be `status, _ = ...`, if you don't care about the 2nd item and just want to toss it aside.

LogosLinuxInstaller.py
```diff
@@ -201,6 +205,17 @@ def parse_args(args, parser):
...
+            if status == 0:
+                config.ACTION = utils.update_to_latest_recommended_appimage
+            elif status == 1:
+                msg.logos_error("The AppImage is already set to the latest recommended.", "info")
+            elif status == 2:
+                msg.logos_error("The AppImage version is newer than the latest recommended.", "info")
+        else:
+            msg.logos_error("The command you requested is disabled. The configured install was not created using an AppImage.")  # TODO: Could we disable these at the arg level instead?
```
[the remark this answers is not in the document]
thw26: None I can think of. Will try

LogosLinuxInstaller.py
```diff
@@ -296,6 +311,9 @@ def main():
     if config.ACTION.__name__ == 'remove_install_dir' or config.ACTION.__name__ == 'utils.check_dependencies': # These actions don't require an install
         logging.info(f"Running function: {config.ACTION.__name__}")
         config.ACTION()
+    elif config.ACTION.__name__ == 'utils.update_latest_appimage':
+        logging.info(f"Running function: {config.ACTION.__name__}")
+        utils.update_to_latest_recommended_appimage()
```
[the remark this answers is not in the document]
thw26: If I do that, PyCharm throws an error is the action not existing.

LogosLinuxInstaller.py
```diff
@@ -183,57 +189,60 @@ def parse_args(args, parser):
...
+                if not utils.file_exists(config.APPIMAGE_FILE_PATH):
+                    raise argparse.ArgumentTypeError(
+                        f"Invalid file path: '{config.APPIMAGE_FILE_PATH}'. File does not exist.")
+                if not utils.check_appimage(config.APPIMAGE_FILE_PATH):
+                    raise argparse.ArgumentTypeError(f"{config.APPIMAGE_FILE_PATH} is not an AppImage.")
+            config.ACTION = action
+            b = True
+            break
+    if not b:
         config.ACTION = run_control_panel
```
n8marti, line 232: I see where you're going with this, and I think it looks pretty good. My one suggestion is that I think we can avoid having to use `b` if we just start with `config.ACTION = None`. Then the check at the end can just be:
```
if config.ACTION is None:
    config.ACTION = run_control_panel
```

LogosLinuxInstaller.py
```diff
@@ -290,24 +299,33 @@ def main():
...
     elif utils.app_is_installed():
         # Run the desired Logos action.
         if config.ACTION is None:
             msg.logos_error("No function given for this subcommand.")
         logging.info(f"Running function: {config.ACTION.__name__}")
-        config.ACTION() # defaults to run_control_panel()
+        config.ACTION()  # defaults to run_control_panel()
     else:
         logging.info(f"Starting Control Panel")
         run_control_panel()
```
n8marti, line 327: The reason for the `if config.ACTION.__name__ == '...'` check is to let certain actions happen even if the app isn't fully installed, such as `remove_install_dir` and `check_dependencies`. So if there are additional actions that fall in that category, then I think we could streamline this by having a list of "pre-install" or "install-agnostic" actions. So we could end up with something like:
```python
install_not_required = [ 'action1', 'action2', 'actionN' ]
if config.ACTION.__name__ in install_not_required:
    logging.info(f"Running function: {config.ACTION.__name__}")
    config.ACTION()
elif utils.app_is_installed():
    # unchanged
else:
    # unchanged
```

wine.py
```diff
@@ -31,7 +31,7 @@ def wait_on(command):
     try:
         # Start the process in the background
         process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
-        msg.cli_msg(f"Waiting on \"{' '.join(command)}\" to finish.", end='')
+        msg.cli_msg(f"Waiting on \"{' '.join(command)}\" to finish…")
         while process.poll() is None:
             msg.logos_progress()
             time.sleep(0.5)
```
n8marti, line 37: The reason for `end=''` here is so that the later `msg.logos_progress()` function can print a "progress" dot every 0.5 sec on the same line as "Waiting on ...". This change has the effect of printing the full message line plus a line return, then printing the dots (maybe only 5 at most) on the next line. Maybe that's what's intended?

thw26: Ah, that explains it. I saw messages in the logs that had two periods which should only have one. I changed this line in order to circumvent "ugly" logs. Now I understand—yes, so the reason I saw two was the hard-coded one and an initial progress dot.

Hmm… I'm not sure the best method here, unless we pass along an additional var to determine whether or not to trigger progress dots… I can see the benefit of just keeping them as is but I can also see the benefit of having both options on hand, and it isn't like we're talking about a major issue… Hmm…

LogosLinuxInstaller.py
thw26: So the main addition to this function is the "disabled" line, in order to account for disabled functions. I'm guessing you are referring to lines 310, 313, and 316.

Those should be easy to fix. Am I understanding this comment properly?

LogosLinuxInstaller.py
thw26: Ah, that's good.

wine.py
thw26: What if I reverted this change and add a 1 second sleep before the timer kicks in?

n8marti (approved):
Visually this look good to me. I don't have time today to test it, though.

n8marti:
These 3 commits get the GUI to a good state, I think. The main changes are described by each commit's message:

1. Use Path object methods to handle symlink setup. The `os` methods are just as good, but since we were already using Path objects in parts of the code, I just decided to make it consistent.
2. The biggest changes were to put the AppImage update into a GUI thread, then run the progress bar while it's happening. This only uses "indeterminate" progress, because that was the simplest way to implement it for now, but we could look toward implementing the download part of the updated AppImage as "determinate" progress down the road.
3. I also threw in PEP8 cleanup for `gui_app.py`. We're both already tempted to do this here and there, and since I was spending so much time in the module, I figured I'd just go ahead and clean it up.

gui_app.py
```diff
@@ -609,6 +750,23 @@ def update_app_button(self, evt=None):
...
+            msg = "This button is disabled. The AppImage is already set to the latest recommended."  # noqa: E501
+        elif status == 2:
+            state = 'disabled'
+            msg = "This button is disabled. The AppImage version is newer than the latest recommended."  # noqa: E501
+        if msg:
+            gui.ToolTip(self.gui.latest_appimage_button, msg)
+        self.clear_message_text()
+        self.stop_indeterminate_progress()
+        self.gui.latest_appimage_button.state([state])
+
```
thw26, line 769: By moving this code here rather than checking it during the creation of the AppImage, the button remains "Enabled" but now on hover says creates a popup saying it is disabled.

Is TK able to handle a dynamic change like this? Can this change be moved to gui_app.py, line 566?

PR #73, opened 2024-03-05
Properly update app button in GUI after install is complete

### LINKED ISSUE #72: GUI doesn't update "Install" button to "Run" after installation.
GUI doesn't update "Install" button to "Run" after installation. The user has to close and re-open the app to see the button changed to "Run Logos".

The solution has to do with somehow signalling the Control Panel window from the Installer window that the installation has finished.

n8marti:
Fixes #72 

Worked correctly when tested with no config file and no existing install.

### DIFF
--- a/gui_app.py
+++ b/gui_app.py
@@ -84,12 +84,13 @@ class Root(Tk):
 
 
 class InstallerWindow():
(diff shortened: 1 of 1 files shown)


thw26 (commented):
Just need to test

PR #78, opened 2024-03-09
GUI Enhancements

### LINKED ISSUE #58: Potential GUI Improvements
- [ ] Modify CSS of installer panel so that the disabled fields are more distinct from the active fields
- [ ] Make the installer panel's get release list run automatically to reduce clicks (?)
- [ ] Make the installer panel's get exe list run automatically (?)

n8marti:
Fixes #58 
- Run wine exe check in thread
- Add default text to drop-down menus (sufficient in lieu of changing background color?)
- Auto-populate release list and exe list once product and version are set

### DIFF
--- a/gui_app.py
+++ b/gui_app.py
@@ -92,14 +92,18 @@ class InstallerWindow():
         self.win.resizable(False, False)
         self.gui = gui.InstallerGui(self.win)
 
-        # Initialize variables from ENV.
-        self.wine_exe = config.WINE_EXE
-
-        # Set other variables to be used later.
+        # Initialize variables.
+        self.flproduct = None  # config.FLPRODUCT
+        self.release_thread = None
+        self.wine_exe = None
+        self.wine_thread = None
+        self.winetricksbin = None
         self.appimage_verified = None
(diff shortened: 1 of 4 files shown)


gui_app.py
```diff
@@ -531,6 +535,9 @@ def update_wine_check_progress(self, evt=None):
         self.gui.progress.config(mode='determinate')
         self.gui.progressvar.set(0)
         self.gui.winevar.set(self.gui.wine_dropdown['values'][0])
+        # FIXME: Probably need better test for failure condition:
+        if len(self.gui.winevar.get()) == 0:
+            self.gui.wine_check_button.state(['!disabled'])
```
thw26, line 540: Could we add a status message if it fails or a tooltip?

thw26 (approved):

gui_app.py
thw26: Perhaps something similar for the Logos release, too

n8marti: Yes, the Logos release check already has a status message. I'm adding one now to the wine binary check.

PR #82, opened 2024-03-15
Update Arch and SteamOS Deps

c-reeder:
Part of #1 

Prevents installation failure on Arch Linux.

Note that it's likely the dependencies will need to be later combed more thoroughly to remove unnecessary parts, but this will at least allow arch users to get past the dependencies step without failure.


c-reeder:
Removed duplicate `bc` and `patch` from SteamOS and `curl`, `gawk`, `bc`, and `libxml2` from Arch

utils.py
```diff
@@ -338,7 +338,10 @@ def get_package_manager():
...
+        else:  # arch
+            config.PACKAGES = "patch wget sed grep cabextract samba print-manager system-config-printer cups-filters nss-mdns foomatic-db-engine foomatic-db-ppds foomatic-db-nonfree-ppds ghostscript glibc samba apparmor libcurl-gnutls winetricks cabextract appmenu-gtk-module lib32-libjpeg-turbo qt5-virtualkeyboard wine giflib lib32-giflib libpng lib32-libpng libldap lib32-libldap gnutls lib32-gnutls mpg123 lib32-mpg123 openal lib32-openal v4l-utils lib32-v4l-utils libpulse lib32-libpulse libgpg-error lib32-libgpg-error alsa-plugins lib32-alsa-plugins alsa-lib lib32-alsa-lib libjpeg-turbo lib32-libjpeg-turbo sqlite lib32-sqlite libxcomposite lib32-libxcomposite libxinerama lib32-libgcrypt libgcrypt lib32-libxinerama ncurses lib32-ncurses ocl-icd lib32-ocl-icd libxslt lib32-libxslt libva lib32-libva gtk3 lib32-gtk3 gst-plugins-base-libs lib32-gst-plugins-base-libs vulkan-icd-loader lib32-vulkan-icd-loader"
```
thw26, line 344: Can you remove the duplicate `cabextract` package here? Also, let's remove `print-manager system-config-printer cups-filters nss-mdns foomatic-db-engine foomatic-db-ppds foomatic-db-nonfree-ppds ghostscript` and `qt5-virtualkeyboard`.

You can also move the remaining `cabextract` to `L9PACKAGES`.

utils.py
```diff
@@ -338,7 +338,10 @@ def get_package_manager():
...
+        if config.OS_NAME == "steamos":  # steamOS
+            config.PACKAGES = "patch wget sed grep gawk cabextract samba bc libxml2 curl print-manager system-config-printer cups-filters nss-mdns foomatic-db-engine foomatic-db-ppds foomatic-db-nonfree-ppds ghostscript glibc samba extra-rel/apparmor core-rel/libcurl-gnutls winetricks cabextract appmenu-gtk-module lib32-libjpeg-turbo qt5-virtualkeyboard wine-staging giflib lib32-giflib libpng lib32-libpng libldap lib32-libldap gnutls lib32-gnutls mpg123 lib32-mpg123 openal lib32-openal v4l-utils lib32-v4l-utils libpulse lib32-libpulse libgpg-error lib32-libgpg-error alsa-plugins lib32-alsa-plugins alsa-lib lib32-alsa-lib libjpeg-turbo lib32-libjpeg-turbo sqlite lib32-sqlite libxcomposite lib32-libxcomposite libxinerama lib32-libgcrypt libgcrypt lib32-libxinerama ncurses lib32-ncurses ocl-icd lib32-ocl-icd libxslt lib32-libxslt libva lib32-libva gtk3 lib32-gtk3 gst-plugins-base-libs lib32-gst-plugins-base-libs vulkan-icd-loader lib32-vulkan-icd-loader"
```
thw26, line 342: If you move `cabextract` to `L9PACKAGES` below, then remove it here, too.

utils.py
thw26: We could also create a new `OPTIONALPACKAGES` config option here to add the printing packages. Any thoughts on this @n8marti?

c-reeder: @thw26 made those changes. Let me know if you spot anything else to be trimmed out.

n8marti: What kinds of features/improvements would be handled by `OPTIONALPACKAGES`? Just printing support? I guess I'd tend to prefer to keep the user's installation steps simpler by not having another choice to make, unless including `OPTIONALPACKAGES` was the default but then there was a simple CLI option to disable them. That would keep the GUI/TUI cleaner.

thw26: Since we are letting @jg00dman handle SteamOS, then let's just keep it as is. We can put in the README's Arch instructions how to get the printing support rather than supporting optional dependencies.

The reason why I ask is that without these, Arch is missing key Logos functionality, i.e., printing from a Logos resource or document. But I'm okay separating that out.

thw26: @c-reeder, can you make a modification to the README file as part of this PR that adds the printing deps to the instructions?

Once you do that, please feel free to squash your commits and to force push, then we can rebase the changes.

c-reeder: @thw26 , Sorry, I'm getting a little jumbled in the back-and-forth. Just so I'm clear..
1. Do you want me to not touch the SteamOS dependencies at all? E.g. leave them as they would have been before this PR? (not even removing duplicates)
2. If so, should I not move the cabextract package to the L9PACKAGES for SteamOS? If that's the case, it seems like rather than just having my Arch/Steam split condition around config.PACKAGE, I would have a top-level condition (so that PACKAGE_MANAGER_COMMAND_*, L9PACKAGES, and BADPACKAGES are also separately set between Arch/Steam).
3. Or do you just mean to keep the PR the way it currently is but your ask is just to add the printing packages back to Steam?

(either way, I'll add the Readme updates)

thw26: 1, 2. Don't touch SteamOS dependencies, but as noted, move cabextract to L9PACKAGES.
3. I'd leave printing in SteamOS and remove from Arch proper.

Perhaps in the future we will consider adding optional dependencies. For now, let's just add the printing dependencies to the README under Arch so that users know what is needed.

Otherwise, squash your commits to a single commit, and all should be well!

thw26:
See discussion in #90 for potential further changes required.

thw26:
`cabextract` should be added back to the main package lists. It is needed by winetricks.

`patch` I believe can be safely removed

PR #89, opened 2024-04-03
Incorporate Unit testing

n8marti:
Unit testing is intended to ensure expected behavior for individual functions. It's not intended to ensure expected behavior of large components or main functionality of the app/script, which would be considered integration testing. But the addition of unit testing should help prevent future bugs.

However, to be effective, new tests would need to be written for any new functions. We can discuss how this might work out in practice.

I've added a command to run the unit tests in the build script and abort the build if the unit tests fail. I think this is a good idea in theory. We'll have to see how it works out in practice.

Unit tests can be run locally if you `pip install coverage`, then run `python -m coverage run -m unittest -b -v`. Note: I had to re-build python 3.12 on my system because my original build didn't include an `sqlite` module needed by `coverage`. So the `ensure-python.sh` script has been modified to make this work correctly.


thw26:
You can drop https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/89/commits/eff4d998ad90375314674627230a600e80817ac1.

n8marti:
dropped redundant commit, rebased on current `main`, and squashed the rest

thw26:
Closing as superseded by https://github.com/FaithLife-Community/OuDedetai/pull/337.

PR #90, opened 2024-04-13
Update Debian and Ubuntu Deps

tbleher:
This fixes two minor issues I found while trying to install on Ubuntu 22.04. Note that `ensure-python.sh` doesn't work on Ubuntu 22.04 yet, since winehq-staging is not found in the repo added to apt. But I think these two bugfixes are still worth merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ NOTE: The following section is WIP.
 ### Install Dependencies
 
 ```
-sudo apt install mktemp patch lsof wget find sed grep gawk tr winbind cabextract x11-apps bc 
+sudo apt install coreutils patch lsof wget findutils sed grep gawk winbind cabextract x11-apps bc binutils
 ```
 
 If using wine from a repo, you must install wine staging. Run:
--- a/utils.py
+++ b/utils.py
@@ -367,7 +367,7 @@ def get_package_manager():
         # IDEA: Switch to Python APT library?
         # See https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/33#discussion_r1443623996  # noqa: E501
         config.PACKAGE_MANAGER_COMMAND_QUERY = "dpkg -l | grep -E '^.i  '"
-        config.PACKAGES = "binutils cabextract fuse wget winbind"
+        config.PACKAGES = "coreutils patch lsof wget findutils sed grep gawk winbind cabextract x11-apps bc binutils fuse3"
         config.L9PACKAGES = ""  # FIXME: Missing Logos 9 Packages
         config.BADPACKAGES = "appimagelauncher"
     elif shutil.which('dnf') is not None:  # rhel, fedora


thw26:
Thank you for the contribution!

Do you know if these package changes apply to all Debian variants?

If so, could you also modify the following lines?

https://github.com/FaithLife-Community/LogosLinuxInstaller/blob/e9ac19d1c1bba731ba65568be6fabbe5791ed632/utils.py#L342-L348

If this is Ubuntu specific, we will need to add a flag for distro. Compare the following:

https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/82/files

thw26 (changes requested):
Initial review. Needs some minor changes.

scripts/ensure-python.sh
```diff
@@ -19,7 +19,7 @@ fi
 # Warn about build deps.
 echo "Warning: You will likely need to install build dependencies for your system."
 echo "e.g. Ubuntu requires: build-essential libreadline-dev tk-dev tcl-dev"
-read -pr "Continue? [Y/n] " ans
+read -r -p "Continue? [Y/n] " ans
```
n8marti, line 22: Good catch. I just used this script a few days ago--not sure how it didn't error on me. I may have modified it after having used it. The "message" is an option argument to `-p`, so it has to follow directly behind it. There's no way this should have worked before, regardless of bash version.

n8marti:
> This fixes two minor issues I found while trying to install on Ubuntu 22.04. Note that `ensure-python.sh` doesn't work on Ubuntu 22.04 yet, since winehq-staging is not found in the repo added to apt. But I think these two bugfixes are still worth merging.

`ensure-python.sh` shouldn't be affected by whether or not winehq-staging package is installed. But if you are talking about the `README.md` instructions, they have you first install winehq's sources file, which is what makes winehq-staging available to apt. However, I see that those instructions are probably incomplete in that there should be a `sudo apt update` command in between the `sudo wget` command and the `sudo apt install` command. `sudo apt update` makes apt look at the new winehq apt source for the first time after adding the the sources file in the earlier step.

n8marti:
> I see that those instructions are probably incomplete in that there should be a `sudo apt update` command in between the `sudo wget` command and the `sudo apt install` command.

This particular oversight has been fixed in [6f8d420](https://github.com/FaithLife-Community/LogosLinuxInstaller/commit/6f8d420a4946defd292cc067f756f25a6c886995)

tbleher:
> Thank you for the contribution!
> 
> Do you know if these package changes apply to all Debian variants?

Yes - should have mentioned this in the commit message. The package names are the same in Debian and Ubuntu.

> If so, could you also modify the following lines?
> 
> https://github.com/FaithLife-Community/LogosLinuxInstaller/blob/e9ac19d1c1bba731ba65568be6fabbe5791ed632/utils.py#L342-L348

I'm a bit confused about the difference between these lists. One list for Debian/Ubuntu currently reads:

> mktemp patch lsof wget find sed grep gawk tr winbind cabextract x11-apps bc

The other reads:

> binutils cabextract fuse wget winbind

I did not find any usage of `patch` in the installer, and `sed`, `bc` and `lsof` also seems unused. For `binutils` I'm not completely sure (since it contains a number of binaries), but I don't see any direct usage.
OK, looking a bit further, `sed` and `binutils` seems to be used by `winetricks`.

I'll upload a patch to unify the two lists, but I guess `patch`, `bc` and `lsof` could be removed? Are these remnants of the old installer, or are they still needed?

tbleher:
> `ensure-python.sh` shouldn't be affected by whether or not winehq-staging package is installed. But if you are talking about the `README.md` instructions, they have you first install winehq's sources file, which is what makes winehq-staging available to apt. However, I see that those instructions are probably incomplete in that there should be a `sudo apt update` command in between the `sudo wget` command and the `sudo apt install` command. `sudo apt update` makes apt look at the new winehq apt source for the first time after adding the the sources file in the earlier step.

Sorry, I guess I mixed things up during testing. Thank you for the correction and the quick fix!

n8marti:
> I did not find any usage of `patch` in the installer, and `sed`, `bc` and `lsof` also seems unused. For `binutils` I'm not completely sure (since it contains a number of binaries), but I don't see any direct usage. OK, looking a bit further, `sed` and `binutils` seems to be used by `winetricks`.

Regarding winetricks' dependencies: This seems like a bit of a can of worms. Here's what I see in the winetricks script:
```shell
$ cat /usr/bin/winetricks | grep -EnB1 '#\s+sudo .* install'
42-# On Ubuntu (20.04 and newer), the following line can be used to install all the prerequisites:
43:#    sudo apt install aria2 binutils cabextract fuseiso p7zip-full pkexec tor unrar unzip wine xdg-utils xz-utils zenity
--
45-# On older Ubuntu, the following line can be used to install all the prerequisites:
46:#    sudo apt install aria2 binutils cabextract fuseiso p7zip-full policykit-1 tor unrar unzip wine xdg-utils xz-utils zenity
--
48-# On Fedora, these commands can be used (RPM Fusion is used to install unrar):
49:#    sudo dnf install https://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm https://download1.rpmfusion.org/nonfree/fedora/rpmfusion-nonfree-release-$(rpm -E %fedora).noarch.rpm
50:#    sudo dnf install binutils cabextract fuseiso p7zip-plugins polkit tor unrar unzip wget wine xdg-utils xz zenity
```
@thw26 What do you think about these? Should we copy them all into our own prereq lists? (Not exactly realistic, since even `wine` is included in that list.) Should we dig a bit deeper and figure out exactly which ones are needed for Logos and leave the rest out? (Maybe this was already done previously before I came along.)

n8marti:
> I'll upload a patch to unify the two lists, but I guess `patch`, `bc` and `lsof` could be removed? Are these remnants of the old installer, or are they still needed?

I would guess that they're remnants from when LogosLinuxInstaller was written in bash and can be removed, but I'll let @thw26 confirm.

thw26:
Several of these, yes, are remnants of the old installer, like `bc` and `lsof`. I think `patch` was used in the old installer for Logos 9. The `winetricks` dependencies add a different concern though—the only way to confirm that is to test.

thw26:
For Ubuntu, should we add a command to add the PPA for Ubuntu users or leave as is? In time, with enough work from upstream, we won't need the dev or staging branch.

n8marti:
> In #95, I note that both wget and curl are deps—we can probably drop curl as our stuff is using python, right? Do we even need wget then?
We're not using `wget`, but winetricks uses it (or alternatively aria2, if it's installed, I think).

thw26:
`cabextract` is used by winetricks

thw26:
Merging this. We can deal with winetricks deps in another issue.

n8marti:
Unfortunately, I just realized that the Ubuntu package dependency `fuse` was changed to `fuse3`. This was probably necessary for newer Ubuntu, but it breaks the dependency check for 20.04, which is still supported until 2025. I've opened a new [issue](#111).

PR #93, opened 2024-04-20
ensure-python.sh: correct shell script

tbleher:
The script aborted since 'read -pr "str" ans' is not valid syntax.


tbleher:
@n8marti Thank you for your reviews in #90. :+1: 

I split out this commit, so it can be merged separately. I have some questions regarding the other commit in #90, which I'll discuss there.

n8marti (approved):
Good catch.

Issue #42, opened 2024-01-10
TUI Improvements


thw26:
- [x] Add Curses file selector in place of curses_get_user_input() for file selections. See #99 and https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/99/commits/fc44b009f91366ec6e67a042c8b04b42bc8b5cd7
- [x] Create a tui_screen.TextScreen (see #99) that shows the list of required packages and show them being scanned/checked off when found, as a form of user feedback, during the dependency checking.
- [x] Add a status line. See #99.
  - When the TUI is running, we cannot see the CLI itself. We may want to consider making the TUI a full app, as in do not return to the CLI but keep the TUI open even on a blank screen, so as not to obscure output.
- [x] Add a progress bar. See #85.
- [x] Rework using curses.wrapper() for error handling. See https://github.com/FaithLife-Community/LogosLinuxInstaller/commit/818358897db2c8f1b95c47212226defc8952adc9.
- [x] Add a progress spinner/wheel to show that we are still working and not stalled. See #99.
- [x] Reduce number of screens by having multiple choices on one??? Probably too painful to generate nicely.
- [x] Implement doupdate instead of refresh

thw26:
Some resources for the status line:

https://stackoverflow.com/questions/46011314/progress-bar-in-python-curses
https://stackoverflow.com/questions/13251299/creating-a-progress-bar-status-in-python-text-console
https://stackoverflow.com/questions/2122385/dynamic-terminal-printing-with-python

thw26:
See also:

https://github.com/wong2/pick
https://github.com/anafvana/pickpack

PR #99, opened 2024-04-26
Add TUI Status Console

thw26:
This PR adds a working TUI console as a partial fulfillment of #42. It involves some major code restructuring in order to be able to do this. Notably, the code is now much more readable and reusable. I'm sure there are further gains to be had in code reuse.

The TUI console is currently given in the top left corner of the screen and shows the three most recent messages. The TUI console is updated separately from the menu window, which is now designed/intended to be replaced with nested menus.

As part of this rework, the keyboard is completely free in its processing, namely, the keyboard does not lock up the program's processing but continues chugging along as it waits for user input. It gracefully handles `Ctrl+C`, and it now also quits when the user hits `ESC`.

---
# Remaining to do before out of draft

- [ ] Test an install all the way through and fix any problems which arise 

---

Further to be done/reviewed/reworked are menus randomly created outside of the TUI or nested menus. This initial commit is focused on restructuring to make this possible.


thw26:
The TUI console works now to get a full installation. Some issues remain:

- [ ] Creates a directory with various contents in it in the dir in which the program is run called `None`. Determine what is running this.
- [ ] When not at a menu, all the info and debug lines appears in the background. Figure out how to silence these so that messages instead appear in the console.
- [ ] Implement a waiting page (?) where no user input can be accepted but the console continues rolling.

thw26:
My current work around for avoiding threads is basically completed at this point. The program works around them now by recalling the appropriate `ensure_*()` when a menu finishes.

The main problem with this is that the install steps are reported and updated wrongly.

---

The proper solution would be to implement threading like in the GUI.

For this I need to learn how to program threads.

thw26:
Accidentally merged this one early by pushing to the wrong branch. I force pushed to the main repo to revert to the previous commit. I opened a new PR: #105.

PR #100, opened 2024-04-28
Fix Install Issues

n8marti:
If an Ubuntu user is running the GUI and needs dependencies installed, the installer will hang because it's waiting fruitlessly for a `sudo` password. The user would only maybe notice this if they start the installer from the terminal, which not a typical use-case for the GUI. Otherwise the prompt is unreachable. This change prioritizes `pkexec` over `sudo`, because `pkexec` will pop up a graphical password prompt. I think this will also give a graphical prompt for a user running the TUI, if there's an available display, otherwise it will fallback to a terminal prompt.

The 2nd change is making sure the TUI checks for "Install dependencies" before checking for a choice that starts with "Install". The original order meant that if the user selected "Install dependencies" from the menu, it actually ran the full installer, since the choice started with "Install".

Both of these changes have been tested on a new Ubuntu 24.04 installation.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -133,6 +133,7 @@ class UrlProps(Props):
 # Set "global" variables.
 def set_default_config():
     get_os()
+    get_superuser_command()
     get_package_manager()
     if config.CONFIG_FILE is None:
         config.CONFIG_FILE = config.DEFAULT_CONFIG_PATH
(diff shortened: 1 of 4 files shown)


utils.py
```diff
@@ -357,7 +357,9 @@ def get_os():
 
 def get_package_manager():
     # Check for superuser command
-    if shutil.which('sudo') is not None:
+    if shutil.which('pkexec') is not None:
+        config.SUPERUSER_COMMAND = "pkexec"
+    elif shutil.which('sudo') is not None:
         config.SUPERUSER_COMMAND = "sudo"
     elif shutil.which('doas') is not None:
         config.SUPERUSER_COMMAND = "doas"
```
thw26, line 365: When I implemented more supported dialogs in Bash, I had a simple flag to check whether we were on GTK or Plasma; the flag allowed me to switch the order of the requested dialog so that GTK would get zenity and Plasma would get kdialog. Though kdialog was never implemented, the change would have made a big difference for making the installer feel integrated.

I think we should add a flag here to check if we are TK or TUI/CLI and swap the order, and that way in TK we don't even request sudo or doas since they would never appear anyways.

thw26 (changes requested):

n8marti:
The last change only checks for `pkexec` if `DIALOG=tk`. This works fine for ubuntu and derivatives. Do other OSes also have `pkexec`? Or does there need to be more OS-specific digging to figure out how to display password prompts in GUI?

thw26 (approved):

thw26:
Do other OSes also have pkexec? Or does there need to be more OS-specific digging to figure out how to display password prompts in GUI?

Yes, I don't know if all do, but Manjaro (my distro) does. I'd assume it's a default of sorts.

Issue #104, opened 2024-05-31
WINE_EXE doesn't get set correctly in Logos_on_Linux.json


Fiyorin:
was trying to install logos on debian 12 but ran into this error: https://pastebin.com/raw/BEkrz7na

looks like WINE_EXE in Logos_on_Linux.json was not set to a valid wine binary, even though i had wine-staging installed before.
so i manually changed the path of WINE_EXE in the config and the installation proceeded

thw26:
I am doing some work in #99 and I believe I came across this while I was working on it and fixed it there.

I will push my latest changes. Would you be willing to test it?

thw26:
The system installations on an Arch system won't be found because their version is greater than 9.0. Wine made some kind of change in 9.0+ that broke our usage.

We need someone to be willing to do a bisect between wine 8.26, or thereabouts, and 9.0+ to figure out what is going on.

See https://github.com/FaithLife-Community/LogosLinuxInstaller/blob/70ffcc090f5d4025733e4753cd47253862563cff/wine.py#L109-L159 for what we are doing.

---

The AppImage is a different story. Is the AppImage not appearing at all in the list, or what error are you seeing?

clawoflight:
If I ever get this working and find the time, I can give the bisection a go :)

The AppImage doesn't appear in the list. In fact, I can never get any wine executable to show up in any way.
I'm at my work machine now, but IIRC it even logs a file not found at some point when I try to force the appimage using the command-line parameter.
In the TUI I simply cannot select anything, it shows the path as ~/Logos*/**/None.

clawoflight:
This is everything useful I get in the terminal:

```
2024-06-05 20:07:37 ERROR: HTTP error occurred: 401
2024-06-05 20:07:37 CRITICAL: Could not get latest release URL.
2024-06-05 20:07:37 CRITICAL: Unable to set LogosLinuxInstaller release without URL.
2024-06-05 20:07:37 ERROR: HTTP error occurred: 401
2024-06-05 20:07:37 CRITICAL: Could not get latest release URL.
2024-06-05 20:07:37 CRITICAL: Unable to set recommended appimage config without URL.

Which Wine AppImage or binary should the script │
                                                                     │ use to install Logos v29.1.0.0022 in            │
                                                                     │ /home/bennett/LogosBible10?                     │
                                                                     │ ┌─────────────────────────────────────────────┐ │
[...]

Issue #102, opened 2024-05-14
For Ubuntu, Install Wine Staging


thw26:
Wine Staging on Ubuntu requires additional install steps besides packages. Add these steps.

PR #105, opened 2024-06-01
Overhaul TUI

### LINKED ISSUE #42: TUI Improvements

thw26:
This PR overhauls the TUI. It involves some major code restructuring in order to be able to do this. Notably, the code is now much more readable and reusable. I'm sure there are further gains to be had in code reuse.

There may be other needs down the road.

Fixed #42, #102, #104


thw26:
Some early feedback:

>So we need a check then in the TUI code to determine someone's dialog version to make sure it is at least a particular version, then we need to change the dialog flag to be a config option…

thw26:
Current conflicts exist because I cherry-picked a commit for #102.

thw26:
Attempting to test a function to test the dialog version and hitting an error when running `./main.py -I`. Hold on review.

thw26:
Fixed the issue. It appears Logos introduced a BOM in their XML data.

thw26:
Fixed `-I`. Now I'm getting errors on my dialog version checker. Progress!

thw26:
And the test is now fixed. Should be ready for review. Again, the conflicts are due to a cherry-pick followed by a rebase.

thw26:
Dropped the XML BOM commit and fixed the cherry-pick/rebase issue.

thw26:
And I have reworked the `dialog` package detection so that it should fallback to curses if the dialog package is outdated. This may break the installer but since I cannot test it at present, I am assuming it will work. The one place I expect it to break is when choosing the install path.

installer.py
```diff
@@ -19,27 +17,11 @@ def ensure_product_choice(app=None):
     logging.debug('- config.FLPRODUCTi')
     logging.debug('- config.VERBUM_PATH')
 
-    if not config.FLPRODUCT:
```
n8marti, line 22: The reason for having the `if not config.VARIABLE` statements at each install step was to be able to skip the step if the variable was already set. Without them each step gets processed in full, even if the user has already set the variable, either in the config file or with env variables.

thw26: I believe I commented those out from an earlier revision, before I had set up proper threading in the TUI. Those should be easy to restore.

thw26:
Most recent push simplifies tui_app.set_tui_menu_options().

thw26:
Most recent push should fix process sharing between main.py and wine.py.

n8marti:
Success! (mostly) I assume the following error is due to ICU files:
image

I'm going to test again with 29.x.

n8marti:
Well, I got the same error, except I got it because the GUI still installs 34.x instead of the 29.x that I selected. That's a separate issue due to the config file's release version for some reason taking precedence over the once chosen in the GUI. See [here](https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/118#issuecomment-2202365819). After removing the config file and trying again it works.

In short, I think we're ready to merge this in.

PR #134, opened 2024-07-22
Update required packages for debian and ubuntu

n8marti:
This simplifies the required packages list for debian, relying on the fact that `binutils` has several of the other previous required packages as dependencies. It also simplifies the list for ubuntu, requiring `fuse3` for all 3 supported releases (20.04, 22.04, 24.04). Some pep8 code cleanup is in here too.


n8marti:
In debian 12 (bookworm) `binutils` pulls in all the previous packages, as I said above. Specifically, here is the relevant part of the dependency tree:
- binutils
  - binutils-common
  - libbinutils
  - binutils-x86-64-linux-gnu
    - libctf-nobfd0
    - libctf0
    - libgprofng0

In debian 11 (bullseye) `binutils` pulls in all but `libgprofng0`, which turns out not to exist in 11. See the [search results](https://packages.debian.org/search?suite=all&section=all&arch=amd64&searchon=names&keywords=libgprofng0) and trying a [direct link](https://packages.debian.org/bullseye/libgprofng0). So this confirms to me that it seems better to go with the top-level package and let it pull in required dependencies.

I think there's enough confirmation to merge this in, then test in a debian 11 VM before the next release.

PR #151, opened 2024-08-02
Add log rotation. Fix #132.

### LINKED ISSUE #132: Add System Log Rotation
The LLI log can grow quite large. Introduce standard log rotation.

This can be done through the logging module. We should probably rotate based on file size.

thw26:
Fixes #132 with gzipped rotation for logs of 10MB in size.

### DIFF
--- a/msg.py
+++ b/msg.py
@@ -1,6 +1,9 @@
+import gzip
 import logging
+from logging.handlers import RotatingFileHandler
 import os
 import signal
+import shutil
 import sys
 import time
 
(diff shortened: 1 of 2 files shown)


thw26:
With this, a 44M log dropped to 3.2M.

n8marti (approved):
Works great! Tested by extending my log file with `truncate` and starting the installer. Also, I guess I had already commented out the only JSON debug call. It's removed now.

PR #152, opened 2024-08-02
144 dependency review

### LINKED ISSUE #144: Too many calls to pkexec during dependency checks
There are at least 4 calls to pkexec during dependency checks. See if they can be reduced to 1 call just for the actual install command. Listing installed packages and downloading new packages shouldn't required elevated privileges (at least not in Ubuntu derivatives). Maybe this isn't true for other distros.

Relatedly, test if winehq repos are actually needed for LLI to run properly.

thw26:
Fixes #144. It also fixes issues with the GUI.

Also reworks how we trigger sudo requests and even stores the user's password in order to reduce the number of requests made to the user.

This PR should bump the release to alpha14.

TODO:

- [x] Verify if TUI password is correct. We need some command we can run and then verify its output. Perhaps ` if sudo whoami == 'root'` or if user ID == 0?
- [x] Verify Python Dialog works
- [x] While coding password testing, I found that the script was not actually installing packages. I added nano as a requirement for Manjaro and then removed it from my system. When running the commands, I found that they reported success but when I checked my system, nano was still not installed. We need to determine what's happening here.
- [x] Need a TK user input prompt that requests the user's password, displaying only a corresponding string of asterisks, that returns the entered password to the app in `system.set_password()`. This requires calling, modifying, removing, etc. the following lines:

https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/152/files#diff-3cbc0fcda69a9da6e615e2a4fe21b03f8ea6face1a58f166387467d0b0ac6762R338-R354
- [x] give user terminal command to install deps in arch and fedora
- [x] update msg.logos_continue_question for if AppImageLauncher is installed
- [x] update msg.logos_error for missing dependencies in arch and fedora


thw26:
We might want to look at using PAM for authentication and the subprocess.run input var for passing user input rather than echo.

Issue #138, opened 2024-07-30
If Completed/Installed Config Exists but Logos EXE is Missing, Program Fails


thw26:
We need code to check if the config is valid, else the program breaks. Ideally we should notify the user if the config can't find the declared install and offer either to install or prompt for the user to check their config.

Make sure config paths are relative so that the install is self-contained and can be moved freely; that is, make sure the config doesn't break if someone moves their install from their initial install dir. cf. https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/13.

n8marti:
I ran into this while testing, and it's not clear what the precise conditions are to make this happen. In my testing I had a proper Logos Dir, etc., except that Logos itself wasn't installed within the wine prefix. The curses UI would just return to the initial screen when choosing either "Install Logos" or "Exit".

Removing the config file was the solution, but it doesn't seem to be related to bad config variables but rather something about the timing of when the config file is looked for when LLI runs. This needs more testing for sure.

thw26:
Somehow this function is probably involved:

https://github.com/FaithLife-Community/LogosLinuxInstaller/blob/b472f4f85443e5ce64f05565df2cc8133c9a4fed/tui_app.py#L525

PR #158, opened 2024-08-24
Config File Issues

### LINKED ISSUE #123: Config File is Not Properly Read or Updated
See:

https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/105#issuecomment-2213818123

https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/118#issuecomment-2202400193

thw26:
This PR will aim to resolve #123 and #138.

I also did a lot of code clean up.

### DIFF
--- a/installer.py
+++ b/installer.py
@@ -140,7 +140,7 @@ def ensure_wine_choice(app=None):
     logging.debug('- config.WINE_EXE')
     logging.debug('- config.WINEBIN_CODE')
 
-    if config.WINE_EXE is None:
+    if utils.get_wine_exe_path() is None:
         network.set_recommended_appimage_config()
         if app:
             utils.send_task(app, 'WINE_EXE')
@@ -152,22 +152,22 @@ def ensure_wine_choice(app=None):
             logging.critical(m)
     else:
         if config.DIALOG == 'curses':
-            app.set_wine(config.WINE_EXE)
+            app.set_wine(utils.get_wine_exe_path())
 
     # Set WINEBIN_CODE and SELECTED_APPIMAGE_FILENAME.
-    m = f"Preparing to process WINE_EXE. Currently set to: {config.WINE_EXE}."
+    m = f"Preparing to process WINE_EXE. Currently set to: {utils.get_wine_exe_path()}."
     logging.debug(m)
-    if config.WINE_EXE.lower().endswith('.appimage'):
-        config.SELECTED_APPIMAGE_FILENAME = config.WINE_EXE
+    if str(utils.get_wine_exe_path()).lower().endswith('.appimage'):
+        config.SELECTED_APPIMAGE_FILENAME = str(utils.get_wine_exe_path())
     if not config.WINEBIN_CODE:
-        config.WINEBIN_CODE = utils.get_winebin_code_and_desc(config.WINE_EXE)[0]  # noqa: E501
+        config.WINEBIN_CODE = utils.get_winebin_code_and_desc(utils.get_wine_exe_path())[0]  # noqa: E501
 
     logging.debug(f"> {config.SELECTED_APPIMAGE_FILENAME=}")
     logging.debug(f"> {config.RECOMMENDED_WINE64_APPIMAGE_URL=}")
     logging.debug(f"> {config.RECOMMENDED_WINE64_APPIMAGE_FULL_FILENAME=}")
     logging.debug(f"> {config.RECOMMENDED_WINE64_APPIMAGE_FILENAME=}")
     logging.debug(f"> {config.WINEBIN_CODE=}")
-    logging.debug(f"> {config.WINE_EXE=}")
+    logging.debug(f"> {utils.get_wine_exe_path()=}")
 
 
 def ensure_winetricks_choice(app=None):
@@ -309,7 +309,7 @@ def ensure_appimage_download(app=None):
     config.INSTALL_STEPS_COUNT += 1
     ensure_sys_deps(app=app)
     config.INSTALL_STEP += 1
-    if config.TARGETVERSION != '9' and not config.WINE_EXE.lower().endswith('appimage'):  # noqa: E501
+    if config.TARGETVERSION != '9' and not str(utils.get_wine_exe_path()).lower().endswith('appimage'):  # noqa: E501
         return
     update_install_feedback(
         "Ensuring wine AppImage is downloaded…",
@@ -345,7 +345,7 @@ def ensure_wine_executables(app=None):
     logging.debug('- wineserver')
 
     # Add APPDIR_BINDIR to PATH.
-    if not os.access(config.WINE_EXE, os.X_OK):
+    if not os.access(utils.get_wine_exe_path(), os.X_OK):
         appdir_bindir = Path(config.APPDIR_BINDIR)
         os.environ['PATH'] = f"{config.APPDIR_BINDIR}:{os.getenv('PATH')}"
         # Ensure AppImage symlink.
@@ -479,7 +479,7 @@ def ensure_wineprefix_init(app=None):
                 f"{config.INSTALLDIR}/data",
             )
         else:
-            if config.WINE_EXE:
+            if utils.get_wine_exe_path():
                 wine.initializeWineBottle()
     logging.debug(f"> {init_file} exists?: {init_file.is_file()}")
 
@@ -556,7 +556,7 @@ def ensure_winetricks_applied(app=None):
             "/t", "REG_SZ",
             "/d", "vista", "/f",
             ]
-        wine.run_wine_proc(config.WINE_EXE, exe='reg', exe_args=exe_args)
+        wine.run_wine_proc(str(utils.get_wine_exe_path()), exe='reg', exe_args=exe_args)
     logging.debug("> Done.")
 
 
--- a/installer.py
+++ b/installer.py
@@ -34,6 +34,9 @@ def ensure_product_choice(app=None):
         else:
             m = f"{utils.get_calling_function_name()}: --install-app is broken"
             logging.critical(m)
+    else:
+        if config.DIALOG == 'curses':
+            app.set_product(config.FLPRODUCT)
 
     if config.FLPRODUCT == 'Logos':
         config.FLPRODUCTi = 'logos4'
@@ -62,6 +65,9 @@ def ensure_version_choice(app=None):
         else:
             m = f"{utils.get_calling_function_name()}: --install-app is broken"
             logging.critical(m)
+    else:
+        if config.DIALOG == 'curses':
+            app.set_version(config.TARGETVERSION)
 
     logging.debug(f"> {config.TARGETVERSION=}")
 
@@ -83,6 +89,9 @@ def ensure_release_choice(app=None):
         else:
             m = f"{utils.get_calling_function_name()}: --install-app is broken"
             logging.critical(m)
+    else:
+        if config.DIALOG == 'curses':
+            app.set_release(config.TARGET_RELEASE_VERSION)
 
     logging.debug(f"> {config.TARGET_RELEASE_VERSION=}")
 
@@ -111,6 +120,9 @@ def ensure_install_dir_choice(app=None):
         else:
             m = f"{utils.get_calling_function_name()}: --install-app is broken"
             logging.critical(m)
+    else:
+        if config.DIALOG == 'curses':
+            app.set_installdir(config.INSTALLDIR)
 
     logging.debug(f"> {config.INSTALLDIR=}")
     logging.debug(f"> {config.APPDIR_BINDIR=}")
@@ -138,6 +150,9 @@ def ensure_wine_choice(app=None):
         else:
             m = f"{utils.get_calling_function_name()}: --install-app is broken"
             logging.critical(m)
+    else:
+        if config.DIALOG == 'curses':
+            app.set_wine(config.WINE_EXE)
 
     # Set WINEBIN_CODE and SELECTED_APPIMAGE_FILENAME.
     m = f"Preparing to process WINE_EXE. Currently set to: {config.WINE_EXE}."
@@ -177,6 +192,9 @@ def ensure_winetricks_choice(app=None):
 
         if not winetricksbin.startswith('Download'):
             config.WINETRICKSBIN = winetricksbin
+    else:
+        if config.DIALOG == 'curses':
+            app.set_winetricksbin(config.WINETRICKSBIN)
 
     logging.debug(f"> {config.WINETRICKSBIN=}")
 
--- a/installer.py
+++ b/installer.py
@@ -155,7 +155,7 @@ def ensure_wine_choice(app=None):
             app.set_wine(utils.get_wine_exe_path())
 
     # Set WINEBIN_CODE and SELECTED_APPIMAGE_FILENAME.
-    m = f"Preparing to process WINE_EXE. Currently set to: {utils.get_wine_exe_path()}."
+    m = f"Preparing to process WINE_EXE. Currently set to: {utils.get_wine_exe_path()}."  # noqa: E501
     logging.debug(m)
     if str(utils.get_wine_exe_path()).lower().endswith('.appimage'):
         config.SELECTED_APPIMAGE_FILENAME = str(utils.get_wine_exe_path())
@@ -257,11 +257,11 @@ def ensure_install_dirs(app=None):
 
     if config.INSTALLDIR is None:
         config.INSTALLDIR = f"{os.getenv('HOME')}/{config.FLPRODUCT}Bible{config.TARGETVERSION}"  # noqa: E501
-        config.APPDIR_BINDIR = f"{config.INSTALLDIR}/data/bin"
 
+    config.APPDIR_BINDIR = f"{config.INSTALLDIR}/data/bin"
     bin_dir = Path(config.APPDIR_BINDIR)
     bin_dir.mkdir(parents=True, exist_ok=True)
-    logging.debug(f"> {bin_dir} exists: {bin_dir.is_dir()}")
+    logging.debug(f"> {bin_dir} exists?: {bin_dir.is_dir()}")
 
     logging.debug(f"> {config.INSTALLDIR=}")
     logging.debug(f"> {config.APPDIR_BINDIR=}")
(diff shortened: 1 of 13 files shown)


thw26:
Should we have a menu on running install that asks if the user wants to use their old config or utilize a new one?

thw26:
After testing, #158 appears to no longer be an issue, but #158 also calls for the Logos config to use relative dirs to the installdir. It looks like the main config var that is impacted by that is the WINE_EXE var.

This is found in the following *app files:

https://github.com/FaithLife-Community/LogosLinuxInstaller/blob/fd807def24bb887823ff5a0c3b7315902484703a/gui_app.py#L435-L446

And it will be here in this PR:

https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/158/commits/c613909d16d80b0cfdfbddfffec766285c9281d0#diff-be7d9ff1bf5cd29e78788cdbdfc6a0e25ae02d0340660e50f175d924f862aa67R520-R525

I think the simple way to make this work would be to check if the WINE_EXE var starts with the INSTALLDIR var, and if so, set it to something like:

```
WINE_EXE = os.path.join(INSTALLDIR, "data/bin/wine-devel_9.12-x86_64.AppImage")
```

thw26:
My most recent testing shows that dynamic path generation is working and so this should now resolve #158 entirely.

But I noticed while working in the GUI that checking the logging status now fails and so does swapping it from enabled to disabled.

thw26:
With this most recent push, I cannot replicate the enable/disable issue, which suggests I fixed some other outlying issues.

I've tested this on Manjaro in both TUI and GUI and all is working.

This is my new config file showing a relative path for WINE_EXE:

```
cat ~/.config/Logos_on_Linux/Logos_on_Linux.json
{
    "BACKUPDIR": null,
    "FLPRODUCT": "Logos",
    "FLPRODUCTi": "logos4",
    "INSTALLDIR": "/home/thwright/LogosBible10",
    "LAST_UPDATED": "2024-08-27T12:33:30",
    "LLI_LATEST_VERSION": "4.0.0-alpha.14",
    "LOGS": null,
    "RECOMMENDED_WINE64_APPIMAGE_URL": "https://github.com/FaithLife-Community/wine-appimages/releases/download/9.12-devel/wine-devel_9.12-x86_64.AppImage",
    "TARGETVERSION": "10",
    "TARGET_RELEASE_VERSION": "35.0.0.0520",
    "WINEBIN_CODE": "Recommended",
    "WINECMD_ENCODING": "437",
    "WINETRICKSBIN": "/usr/bin/winetricks",
    "WINE_EXE": "data/bin/wine-devel_9.12-x86_64.AppImage",
    "current_logos_version": "35.0.0.0520"
}
```

n8marti:
This is what I'm getting:

image

Now I see from the scattered exception text that it must be related to the winetricks variable ending up as "Download" somewhere along the way.

Here's the config file:
```
$ cat .config/Logos_on_Linux/Logos_on_Linux.json
{
    "BACKUPDIR": null,
    "FLPRODUCT": "Logos",
    "FLPRODUCTi": "logos4",
    "INSTALLDIR": "/home/nate/LogosBible10",
    "LAST_UPDATED": "2024-08-28T07:07:58",
    "LLI_LATEST_VERSION": "4.0.0-alpha.14",
    "LOGS": null,
    "RECOMMENDED_WINE64_APPIMAGE_URL": "https://github.com/FaithLife-Community/wine-appimages/releases/download/9.12-devel/wine-devel_9.12-x86_64.AppImage",
    "TARGETVERSION": "10",
    "TARGET_RELEASE_VERSION": "34.1.0.0009",
    "WINEBIN_CODE": "Recommended",
    "WINECMD_ENCODING": "850",
    "WINETRICKSBIN": "/home/nate/LogosBible10/data/bin/winetricks",
    "WINE_EXE": "/home/nate/LogosBible10/data/bin/wine-devel_9.12-x86_64.AppImage",
    "current_logos_version": null
}

```

thw26:
So we use either the winetricks path or the word download in order to know what we need to do to get winetricks installed. What that would suggest is that somewhere the keyword "Download" isn't properly triggering a download but instead is being passed as the path. Hmm…

thw26:
Also seeing this error:

```
2024-08-29 08:31:12 ERROR: Curses error in run_control_panel(): addwstr() returned ERR
  File "/home/thwright/Dev/LogosLinuxInstaller/./main.py", line 442, in <module>
    main()
  File "/home/thwright/Dev/LogosLinuxInstaller/./main.py", line 419, in main
    config.ACTION()
  File "/home/thwright/Dev/LogosLinuxInstaller/./main.py", line 304, in run_control_panel
    raise e
  File "/home/thwright/Dev/LogosLinuxInstaller/./main.py", line 292, in run_control_panel
    curses.wrapper(tui_app.control_panel_app)
  File "/usr/lib/python3.12/curses/__init__.py", line 94, in wrapper
    return func(stdscr, *args, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thwright/Dev/LogosLinuxInstaller/tui_app.py", line 708, in control_panel_app
    TUI(stdscr).run()
  File "/home/thwright/Dev/LogosLinuxInstaller/tui_app.py", line 251, in run
    self.display()
  File "/home/thwright/Dev/LogosLinuxInstaller/tui_app.py", line 216, in display
    self.console.display()
  File "/home/thwright/Dev/LogosLinuxInstaller/tui_screen.py", line 92, in display
    self.stdscr.addstr(2 + i, 2, f"{message}")
_curses.error: addwstr() returned ERR
```

It happens at this point in the log:

```
tail -n50 ~/.local/state/Logos_on_Linux/Logos_on_Linux.log
2024-08-29 08:31:04 DEBUG: sed status: Installed
2024-08-29 08:31:04 DEBUG: Checking for: grep
2024-08-29 08:31:04 DEBUG: 'grep' installed: grep                                         3.11-1                            core      911.3 kB
2024-08-29 08:31:04 DEBUG: grep status: Installed
2024-08-29 08:31:04 DEBUG: Checking for: gawk
2024-08-29 08:31:04 DEBUG: 'gawk' installed: gawk                                         5.3.0-1                           core      3.5 MB
2024-08-29 08:31:04 DEBUG: gawk status: Installed
2024-08-29 08:31:04 DEBUG: Checking for: cabextract
2024-08-29 08:31:04 DEBUG: 'cabextract' installed: cabextract                                   1.11-2                            extra     73.9 kB
2024-08-29 08:31:04 DEBUG: cabextract status: Installed
2024-08-29 08:31:04 DEBUG: Checking for: samba
2024-08-29 08:31:04 DEBUG: 'samba' installed: samba                                        4.20.4-1                          extra     65.0 MB
2024-08-29 08:31:04 DEBUG: samba status: Installed
2024-08-29 08:31:04 DEBUG: Checking for: bc
2024-08-29 08:31:04 DEBUG: 'bc' installed: bc                                           1.07.1-5                          extra     176.1 kB
2024-08-29 08:31:04 DEBUG: bc status: Installed
2024-08-29 08:31:04 DEBUG: Checking for: libxml2
2024-08-29 08:31:04 DEBUG: 'libxml2' installed: libxml2                                      2.13.3-1                          core      4.0 MB
2024-08-29 08:31:04 DEBUG: libxml2 status: Installed
2024-08-29 08:31:04 DEBUG: Checking for: curl
2024-08-29 08:31:04 DEBUG: 'curl' installed: curl                                         8.9.1-2                           core      2.0 MB
2024-08-29 08:31:04 DEBUG: curl status: Installed
2024-08-29 08:31:04 DEBUG: Packages status: {'patch': 'Installed', 'wget': 'Installed', 'sed': 'Installed', 'grep': 'Installed', 'gawk': 'Installed', 'cabextract': 'Installed', 'samba': 'Installed', 'bc': 'Installed', 'libxml2': 'Installed', 'curl': 'Installed'}
2024-08-29 08:31:07 DEBUG: packages to check: []
2024-08-29 08:31:07 DEBUG: Packages status: {}
2024-08-29 08:31:07 DEBUG: No missing or conflicting dependencies found.
2024-08-29 08:31:07 DEBUG: Performing pre-install dependencies…
2024-08-29 08:31:07 DEBUG: No pre-install dependencies required.
2024-08-29 08:31:12 DEBUG: 'libfuse' found at '/usr/lib/libfuse3.so.3.16.2'
2024-08-29 08:31:12 INFO: * libfuse is installed!
2024-08-29 08:31:12 DEBUG: fuse_lib_installed=True
2024-08-29 08:31:12 DEBUG: No missing packages detected.
2024-08-29 08:31:12 DEBUG: No conflicting packages detected.
2024-08-29 08:31:12 DEBUG: Performing post-install dependencies…
2024-08-29 08:31:12 DEBUG: No post-install dependencies required.
2024-08-29 08:31:12 DEBUG: No dependency install required.
2024-08-29 08:31:12 DEBUG: > Done.
2024-08-29 08:31:12 DEBUG: Install step 11 of 23
2024-08-29 08:31:12 INFO: Progress: 48%
2024-08-29 08:31:12 INFO: Ensuring wine AppImage is downloaded…
2024-08-29 08:31:12 INFO: 'wine-devel_9.12-x86_64.AppImage' exists in /home/thwright/Downloads.
2024-08-29 08:31:12 DEBUG: Checking /home/thwright/LogosBible10 for wine-devel_9.12-x86_64.AppImage.
2024-08-29 08:31:12 DEBUG: Checking /home/thwright/Dev/LogosLinuxInstaller for wine-devel_9.12-x86_64.AppImage.
2024-08-29 08:31:12 DEBUG: Checking /home/thwright/Downloads for wine-devel_9.12-x86_64.AppImage.
2024-08-29 08:31:12 INFO: wine-devel_9.12-x86_64.AppImage exists in /home/thwright/Downloads. Verifying properties.
2024-08-29 08:31:12 INFO: Verifying /home/thwright/Downloads/wine-devel_9.12-x86_64.AppImage…
2024-08-29 08:31:12 DEBUG: Comparing size of https://github.com/FaithLife-Community/wine-appimages/releases/download/9.12-devel/wine-devel_9.12-x86_64.AppImage and /home/thwright/Downloads/wine-devel_9.12-x86_64.AppImage.
2024-08-29 08:31:12 DEBUG: Getting headers from https://github.com/FaithLife-Community/wine-appimages/releases/download/9.12-devel/wine-devel_9.12-x86_64.AppImage.
2024-08-29 08:31:12 DEBUG: Starting new HTTPS connection (1): github.com:443
2024-08-29 08:31:12 ERROR: Curses error in run_control_panel(): addwstr() returned ERR
```

thw26:
Got this error while running:

```
2024-08-29 09:04:39 ERROR: Error running '/usr/bin/winetricks -q settings renderer=gdi': 1
```

n8marti:
still getting the same "Download" error that I mentioned [above](https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/158#issuecomment-2315158619). This is starting from a clean ubuntu 20.04 system and installing with the [most recent binary](https://github.com/FaithLife-Community/test-builds/releases/download/fix-123-reading-config-file-980a932/LogosLinuxInstaller) built in the test-builds repo.

thw26:
I have not been able to replicate on my main machine, so I may need to start a VM to test.

n8marti:
I got a successful install with newest binary and *no* existing config file, but it still shows "Install..." in the curses window rather than "Run...". Still getting the `WINETRICKSBIN = 'Download'` issue, but trying to trace it down.

thw26:
My work in #164 introduces a consistently refreshing TUI main menu; it might resolve the Install/Run issue. The only thing I can think otherwise is to evaluate why it might be failing to detect the install and thereby fail to update the menu list. I'm guessing some check is failing, as a successful TUI install has updated the menu for me, to my knowledge, every time.

n8marti:
Trying to sort out the `WINETRICKSBIN = 'Download'` problem, and here is the logic I'm following:
- my existing test config file defines `WINETRICKSBIN` as `/home/<user>/LogosBible10/data/bin/winetricks`
- `installer.py` checks for existing `config.WINETRICKSBIN`; if it is not `None` and if using `curses`, then `app.set_winetricksbin(config.WINETRICKSBIN)` called
- `tui_app.set_winetricksbin` checks if the argument starts with "Download" (it doesn't); then `utils.get_winetricks_options()` is called and the 1st result ("Download") is set as `config.WINETRICKSBIN`
- and off we go with `config.WINETRICKSBIN = 'Download'` rather than the pre-defined value in `config.WINETRICKSBIN`

So I think what should happen is a modification to `tui_app.set_winetricksbin`. Here's the current function:
```python
    def set_winetricksbin(self, choice):
        if choice.startswith("Download"):
            self.tricksbin_q.put("Download")
        else:
            winetricks_options = utils.get_winetricks_options()
            config.WINETRICKSBIN = winetricks_options[0]
            self.tricksbin_q.put(config.WINETRICKSBIN)
        self.menu_screen.choice = "Processing"
        self.tricksbin_e.set()
```
but nowhere does it just accept existing `config.WINETRICKSBIN` as-is if it's already defined in the config file. So maybe it should be more like this?:
```python
    def set_winetricksbin(self, choice):
        if config.WINETRICKSBIN is not None:
            self.tricksbin_q.put(config.WINETRICKSBIN)
        elif choice.startswith("Download"):
            self.tricksbin_q.put("Download")
        else:
            winetricks_options = utils.get_winetricks_options()
            config.WINETRICKSBIN = winetricks_options[0]
            self.tricksbin_q.put(config.WINETRICKSBIN)
        self.menu_screen.choice = "Processing"
        self.tricksbin_e.set()
```
There are other ways to solve this; e.g. maybe changing the function in `installer.py`, but I leave that up to @thw26 to figure out because I don't know if there might be other TUI-related ramifications because of how this gets resolved.

thw26:
This is the installer.py for winetricks:

```
    if config.WINETRICKSBIN is None:
        # Check if local winetricks version available; else, download it.
        config.WINETRICKSBIN = f"{config.APPDIR_BINDIR}/winetricks"

        if app:
            utils.send_task(app, 'WINETRICKSBIN')
            if config.DIALOG == 'curses':
                app.tricksbin_e.wait()
            winetricksbin = app.tricksbin_q.get()
        else:
            m = f"{utils.get_calling_function_name()}: --install-app is broken"
            logging.critical(m)

        if not winetricksbin.startswith('Download'):
            config.WINETRICKSBIN = winetricksbin
    else:
        if app:
            if config.DIALOG == 'curses':
                app.set_winetricksbin(config.WINETRICKSBIN)
                app.tricksbin_e.wait()

    logging.debug(f"> {config.WINETRICKSBIN=}")
```

As you can see, the problem is that we never actually trigger set_winetricksbin as a None rule because we define it before we proceed. Looking over to think what is the best way to proceed.

thw26:
This is the other part of the issue. In installer, we split the steps in two: get the user choice and process it later. This is already happening partially in the TUI code. Without also changing the GUI, we need to make sure that we pass along Download.

```
def ensure_winetricks_executable(app=None):
    config.INSTALL_STEPS_COUNT += 1
    ensure_wine_executables(app=app)
    config.INSTALL_STEP += 1
    update_install_feedback(
        "Ensuring winetricks executable is available…",
        app=app
    )

    if not os.access(config.WINETRICKSBIN, os.X_OK):
        tricksbin = Path(config.WINETRICKSBIN)
        tricksbin.unlink(missing_ok=True)
        # The choice of System winetricks was made previously. Here we are only
        # concerned about whether the downloaded winetricks is usable.
        msg.logos_msg("Downloading winetricks from the Internet…")
        system.install_winetricks(
            tricksbin.parent,
            app=app
        )
    if config.WINETRICKSBIN:
        logging.debug(f"> {config.WINETRICKSBIN} is executable?: {os.access(config.WINETRICKSBIN, os.X_OK)}")  # noqa: E501
    return 0
```

thw26:
I think this is a simpler solution. In this case, we now just check if `app` is set or not. Then if `app` is set, we check if the config is None or not. If it is None, we send a task to get a new value. Only in this new situation do we then set config.WINETRICKSBIN to a new value. Otherwise, we keep the old value and we will eventually progress to the ensure winetricks is exectuable step.

```
def ensure_winetricks_choice(app=None):
    config.INSTALL_STEPS_COUNT += 1
    ensure_wine_choice(app=app)
    config.INSTALL_STEP += 1
    update_install_feedback("Choose winetricks binary…", app=app)
    logging.debug('- config.WINETRICKSBIN')

    if app:
        if config.WINETRICKSBIN is None:
            utils.send_task(app, 'WINETRICKSBIN')
            if config.DIALOG == 'curses':
                app.tricksbin_e.wait()
            config.WINETRICKSBIN = app.tricksbin_q.get()
    else:
        m = f"{utils.get_calling_function_name()}: --install-app is broken"
        logging.critical(m)

    logging.debug(f"> {config.WINETRICKSBIN=}")
```

installer.py
```diff
@@ -392,7 +383,7 @@ def ensure_winetricks_executable(app=None):
         app=app
     )
 
-    if not os.access(config.WINETRICKSBIN, os.X_OK):
+    if config.WINETRICKSBIN == 'Download' or not os.access(config.WINETRICKSBIN, os.X_OK):
```
thw26, line 386: I think we need something else here. If the user provides a preset config and we find that it is not executable, we override it rather than attempting to correct it by making it executable or asking the user what to do. What do you think?

n8marti: I feel uneasy about this change because surely the `WINETRICKSBIN` variable itself should never actually be "Download", right? Rather, if the UI defaults to "Download" while preparing the install, then it seems like *it* should set the correct subsequent `WINETRICKSBIN` value rather than the `install` module having to react to that placeholder value.

Or, coming from another direction, the variable *could* simply be just "Download" or "System", rather than giving a file path. Is it really that useful to allow an arbitrary path for `WINETRICKSBIN`? If it's "Download", then the script would just interpret the location as `INSTALLDIR/data/bin/winetricks`; if it's "System", then the script finds its path with `shutil.which('winetricks')`. That would simplify the code a decent amount, right? I know it would require reworking a few other things, but maybe a better long-term solution for easier maintainability?

thw26: I was thinking over this one and I went with the smallest hammer (get existing method working).

I agree this needs to be redone. Do you want that redone in this PR or a new one?

n8marti: I've created Issue #170  to address this, rather than forcing it into this PR.

n8marti:
I'm in the middle of testing, and things are looking pretty good so far. Here's what I've noticed at this point:
- overall things seem to work correctly both with and without an existing config file! :tada: 
- there's a regression during the actual installation of the Logos MSI: it fails when using the `-P` (passive MSI install) flag (I use this in testing all the time to speed it up, but most users probably aren't going to use it)
- there's a GUI regression in that picking a Logos version other than the top one auto-reverts back to the top one

I'm going to look into them a little deeper next.

PR #160, opened 2024-08-27
Fix #127

### LINKED ISSUE #127: Change Application Folder Names
Currently we store our config in `~/.config/Logos_on_Linux` and our logs in `~/.local/state/Logos_on_Linux`. I'd like to suggest we change `Logos_on_Linux` to FaithLife_Community. Then underneath these we could have different log files, such as for installs, updates, etc.

thw26:
Fixes #127.

Should we provide some kind of automatic migration process?

### DIFF
--- a/config.py
+++ b/config.py
@@ -26,8 +26,8 @@ extended_config = {
     'DEBUG': False,
     'DELETE_LOG': None,
     'DIALOG': None,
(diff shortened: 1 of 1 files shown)


n8marti:
I'd say that if we were changing the config and log folders *after* have a stable release, then it might be wise to have a migration step. Otherwise I think people running `alpha` releases should be content with having to manually rename the folders, or copy over the files, or just have to start with a clean slate. My 2 cents, anyway

n8marti (approved):

Issue #147, opened 2024-08-01
Utilize an Abstract Class for GUI, TUI, and CLI


thw26:

n8marti:
I think this is a worthwhile idea. For the GUI nearly everything happens in a background thread, and the GUI window is the about the only thing that happens in the main thread.
- Each background thread basically has its own queue (if any information needs to be passed back to the GUI), its own event, and its own method to handle that event.
- The GUI starts the thread and its associated function.
- When the function is done it puts any needed information into the queue and then generates the event.
- When the event is generated in the thread it calls the associated method in the GUI, which continues the processing.
- If information is needed from the background thread, then this GUI method looks for that in the queue.

So if we use standardized names for the common features (progress indicators, status messages, etc.), then we should be able to streamline the code that uses these features.

thw26:
A way to begin this would be to move our event and thread and queue code to its own module; this would let us treat it more like an API rather than a dialog-specific solution.

thw26:
With the CLI code merged, we now have a minimal example of a Queue/Event model. We can reuse this to bring it to the TUI and GUI and thereby simplify their code greatly.

PR #161, opened 2024-08-28
Campaign to Save the CLI

### LINKED ISSUE #128: `--install-app` Optarg is Broken
Following the TUI overhaul (#105), the `--install-app` optarg has broken. The program is unable to bypass the event wake ups.

~~We likely need to add in a series of `else:` statements to the `ensure_*()` in order to do an automatic or by-config install. We may want to review the removed code.~~ This should be done in #158. This should be retested with #158.

Alternatively, we can drop the optarg but then we remove automatic/unattended installs.

Alternatively, we can make the flag launch the TUI app, but then we remove automatic/unattended installs..

cf. #2

thw26:
Fixes #128. Requires #158.

I should clarify my aim is to get it working and to make it pretty later (cf. #172).

Part of my goal in this PR is to make #147 potentially easier. The CLI is intentionally aiming to be a highly simplified event/queue manager, thus I aim to clear the event's status frequently in order to reuse it later. The TUI is, to my knowledge, never waiting on more than one event at a time. Therefore we have not need to have multiple events. If we do have some place where we are waiting on multiple, then we can have two different ones at that time. This would make the code much simpler, leaner, and easier to reuse.

What I imagine therefore, from this PR, is extracting all the event/queue code from the TUI, the CLI, and the GUI, into its own class module.

### DIFF
--- a/control.py
+++ b/control.py
@@ -211,6 +211,7 @@ def copy_data(src_dirs, dst_dir):
 
 def remove_install_dir():
     folder = Path(config.INSTALLDIR)
+    # FIXME: msg.cli_question needs additional arg
     if (
         folder.is_dir()
         and msg.cli_question(f"Delete \"{folder}\" and all its contents?")
--- a/cli.py
+++ b/cli.py
@@ -1,0 +1,52 @@
+import logging
+import threading
+import queue
+
+import config
+import installer
+import msg
+import utils
+
+
+class CLI:
+    def __init__(self):
+        self.running = True
+        self.choice_q = queue.Queue()
+        self.input_q = queue.Queue()
+        self.event = threading.Event()
+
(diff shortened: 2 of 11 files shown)


control.py
thw26, line 1: This is fixed in #182

n8marti:
`heavy_wineserver_wait()` was only being used after running `winetricks` commands, and after adding debug logging there don't seem to be any processes still using the `WINEPREFIX` when `winetricks` commands finish and `wineserver_wait()` is run. I suggest we keep the debug logging for the time being, that way if anyone has problems b/c of no longer using `heavy_wineserver_wait()` we'll be able to see if it's due to the `WINEPREFIX` still being accessed by something.

n8marti (approved):

Issue #153, opened 2024-08-02
Store Wine Debug Log in a Separate File


thw26:
When running the TUI I don't get to see the Wine debug log. To see it I have to run `-D` which kills the TUI. Storing this in its own log seems most appropriate.

Issue #137, opened 2024-07-27
Track Logos PID


thw26:
If we track the main PID for Logos once it is running, we can put additional safeguards in to make sure commands can't be run while Logos is running. We can also add commands like stop Logos/restart Logos.

This may be best accomplished by making a system.popen_command, like system.run_command.

thw26:
cf. #164

thw26:
This won't work unless we use a series of grep commands to find the Logos PID. When Logos launches, it creates its splash screen. As soon as that loads, it creates the login window, thus changing PID, and again changing PID on logging in.

PR #164, opened 2024-08-31
Add system popen command

thw26:
This begins to resolve a TODO item in system.py. The goal is to replace those instances of system.run_command() with system.popen_command() wherever appropriate. This PR lays the foundation.


thw26:
I pushed my most recent working code, which I believe should work, but basically we lose the PID after launch and thus we lose ability to track it. See comment:

https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/137

thw26:
My recent push resolves the ability to detect whether Logos is running or not. To do:

- [ ] Log in to Logos to verify the .exe is the same
- [ ] The current detection code in tui_app, lines 622–625, are not able to differentiate between Logos running and not running. Rather, once the key exists in the dictionary, it is always detected as running. This needs to be refined.

thw26:
After adding the state manager, there has been a regression. I need to figure out why stopping doesn't work again.

- [ ] Resolve stop not working
- [ ] Add GUI portion

thw26:
The only thing this PR is missing is any required updates to the GUI.

thw26:
I added a commit to fix #153. I also changed utils.stopwatch to accept floats and reduced the TUI app's monitor refresh rate from 3 to 2.5seconds. CPU usage still seems reasonable.

n8marti (approved):
This has been tested as a script and binary, both GUI and TUI. I think it's ready to be merged into `main`.

PR #165, opened 2024-08-31
Fix #148

### LINKED ISSUE #148: Switch to a dictionary for Wine version checking
https://github.com/FaithLife-Community/LogosLinuxInstaller/blob/4110a2fc17a3477a7612301d1b078702629a83b9/wine.py#L114-L172

Our current test is convoluted and will only grow worse. We should switch to a dict for determining bad versions.

thw26:
Fix #148. My goal is to make the code more readable/usable/extensible (for future wine versions).

The goal was to get rid of the lengthy if/else check. Short of just keeping a list of known bad versions for x.0–x.26, I thought a rules based approach would be better, i.e., making what was explicit is now contained in a reusable function.

I haven not noticed any regressions with this change.

### DIFF
--- a/wine.py
+++ b/wine.py
@@ -118,22 +118,68 @@ def get_wine_release(binary):
         return False, f"Error: {e}"
 
 
-def check_wine_version_and_branch(release_version, test_binary):
+def check_wine_rules(wine_release, release_version):
     # Does not check for Staging. Will not implement: expecting merging of
     # commits in time.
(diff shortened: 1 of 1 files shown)


n8marti (approved):
I don't have a quick way to check which Logos versions work with which wine versions and releases, so I'm going to assume you've done at least some spot-checking on that and approve the PR.

PR #179, opened 2024-09-22
Correct ensure-venv command

jimbob88:
Corrects spelling in "scrips" -> "scripts" and a few grammatical mistakes.

Yours in Christ,
James


n8marti (approved):
These changes are fine. Some good clarifications in wording and punctuation, but especially correcting the misspelling of "scripts". It's not necessary to use sequential numbers in GitHub's Markdown syntax, and it can even be helpful to just use a series of "1"s, that way if items are added or removed in the middle of the list there's no need to go back and renumber them. But I can see how this could be confusing if someone reads the README file in plain text.

PR #183, opened 2024-09-28
Fix #17. Add Alpine Support

### LINKED ISSUE #17: Add APK Support
https://github.com/ferion11/LogosLinuxInstaller/issues/213

thw26:
Fix #17.

### DIFF
--- a/ou_dedetai/system.py
+++ b/ou_dedetai/system.py
@@ -320,7 +320,7 @@ def get_package_manager():
         )
         config.L9PACKAGES = ""  # FIXME: Missing Logos 9 Packages
         config.BADPACKAGES = ""  # appimagelauncher handled separately
-    elif shutil.which('zypper') is not None:  # manjaro
+    elif shutil.which('zypper') is not None:  # opensuse
         config.PACKAGE_MANAGER_COMMAND_INSTALL = ["zypper", "--non-interactive", "install"]  # noqa: E501
         config.PACKAGE_MANAGER_COMMAND_DOWNLOAD = ["zypper", "download"]  # noqa: E501
         config.PACKAGE_MANAGER_COMMAND_REMOVE = ["zypper", "--non-interactive", "remove"]  # noqa: E501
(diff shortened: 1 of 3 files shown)


thw26:
Currently errors when running the binary:

```
Error relocating /tmp/_MEIefOMjg/libz.so.1: crc32_z: symbol not found
```

This appears to be related to the use of muslc instead of glibc.

https://github.com/JonathonReinhart/staticx/issues/143

thw26:
I checked this on an Alpine machine. I've gotten further with some necessary commands for fuse, but fuse mounts are still not being recognized.

thw26:
Because we use pkexec sh -c, bash needs to be added as a dep. There is a permission error in Alpine that appears to be caused by something not set in:

- setup-interfaces
- setup-apkrepoa
- apk update
- setup-desktop

thw26:
Okay. This took a bit of doing, and even now it isn't perfect, but I have Logos running on it.

There is something wrong with the AppImage support and I cannot get it to work.

Alpine is added to the list of bad OSes alongside Arch and Fedora.

After doing a manual install at the pkexec prompt for the preinstall commands (we need someway to determine if these are run), for some reason it still won't mount.

Instead I switched to the edge repository for Wine Dev 9.17 as the stable repo only has 9.0 at present.

I then had to write /usr/bin/wine to the config in order to override our wine branch check.

This then bypassed the wine choice in the installer.

After this, it was able to install, though it seemed to blaze through the winetricks steps which makes me think they didn't work.

On running Logos, it causes a critical error that crashes. This was because of the wine branch checking in logos.py. I got around this by adding in an if/else that defaults to trusting the user's config if we can't verify that the branch actually is bad. This probably has some kind of unintended consequence.

Once I did all this, I was then able to run Logos and I got to the login screen.

thw26:
I am requesting a review on this as technically it works, but we need to do some testing to see if my workarounds don't cause additional breakage.

The reason why Alpine will be supported/not supported is that it is waiting on additional package updates.

Also, ideally we would have some kind of check that could determine if AppImage support is available, and if not, we do not list the AppImage as an option.

Unfortunately, what that would mean is that there are no working options for Alpine (until stable's wine version is updated) and a user would meet a blank installer on opening.

But based on testing, the distro's install now works when conditions are met on it, so our support is technically there, just not practically yet, unless the user is on edge.

thw26:
Changing `fuse3` to `fuse` fixed the appimage issues.

Now there's just the oddity of the postinstall commands not working. When run, they error out the installer, but I have absolutely no idea why.

I'd appreciate a second pair of eyes on it, but otherwise, if the commands are run outside of the installer, the install works, and thus Alpine is listed under the bad_os.

ou_dedetai/system.py
```diff
@@ -682,7 +714,7 @@ def install_dependencies(packages, bad_packages, logos9_packages=None, app=None)
             app.root.event_generate('<<StartIndeterminateProgress>>')
         msg.status("Installing dependencies…", app)
         final_command = [
-            f"{config.SUPERUSER_COMMAND}", 'sh', '-c', "'", *command, "'"
+            f"{config.SUPERUSER_COMMAND}", 'sh', '-c', '"', *command, '"'
```
ctrlaltf24, line 717: Not sure about this, would be easier to maintain this as single quotes (to stop any weird expansions)

thw26: I found that I needed this change to get this to work properly; there was an error I can't recall caused by the single quotes.

ctrlaltf24: This command would need to switch to double quotes if this was single https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/183/files#diff-2c221b5dbbbc57e12ff8bd0cd8ca89e953d82f3e9a5356644bde9779e47ea40dR568

PR #184, opened 2024-09-28
Add OpenSUSE Support

### LINKED ISSUE #16: Add Zypper Support
https://github.com/ferion11/LogosLinuxInstaller/issues/214

thw26:
Fix #16.


thw26:
The only issue I'm seeing on an OpenSUSE Leap machine is that it has trouble finding the patch package when it is already installed.

n8marti (approved):
Not sure how it would confirm the other packages but not `patch`, but I see no reason to hold this up over that.

PR #192, opened 2024-10-06
fix: ensure app is set before attempting to operate on it

ctrlaltf24:
I don't know of any code paths to trigger this error, however app may be uninitiated at this point


thw26:
Thank you for all your contributions and issue reporting!

The hypothetical error could only be triggered from the CLI at present; following #161 it should be impossible.

That being said, this seems like a simple change that ensures the program only executes if a UI is found. Merging.

Issue #171, opened 2024-09-14
Program Icon Appears as a Gear on GNOME Desktop Bar


thw26:

jimbob88:
I am not getting this issue on Debian GNU/Linux 12 (bookworm), Gnome 43.9, X11 [verbum + logos 10]:

![image](https://github.com/user-attachments/assets/4d70bfbd-1d9a-4a2e-b0f9-df1822ecd363)

Using build on https://github.com/FaithLife-Community/LogosLinuxInstaller/commit/f02e0133670040337b0351de7582034f827e10ba

n8marti:
This definitely happens if Logos is installed using the script because there's no associated `.desktop` file. But I'm not seeing this either after installing using the binary. @thw26 can you reproduce this after installing via the binary?

thw26:
Is this the launcher's icon or our program's icon?

I'll need time to set up a dev environment, as all my dev environments run Plasma.

PR #198, opened 2024-10-16
convert to pyproject

### LINKED ISSUE #185: Add Application to PIP

n8marti:
Fixes #185 

Restructure the repo so that the app can be installed with pip. This requires some changes to dev workflow:
- `requirements.txt` has been replaced with `pyproject.toml` dependencies. So to install dependencies:
  - `python3 -m pip install .  # installs required runtime dependencies`
  - `python3 -m pip instsall .[build]  # installs runtime & optional build dependencies (e.g. PyInstaller)`
- The app can't be run with `./logos_on_linux/main.py` because it will cause a relative import error. The alternatives are:
  - `python3 -c 'from logos_on_linux import main; main.main() --help`
  - `./scripts/run_app.py --help  # this script is also needed so that PyInstaller can build the app correctly`
  - `logos-on-linux --help  # if installed with pip, which would have to be re-run every time a file is changed`

Still to do:
- [x] determine app name (placeholder is `Logos on Linux`)
- [x] determine python package name (if different from app name; placeholder is `logos_on_linux`) 
- [x] determine executable name (if different from app name; placeholder is `logos-on-linux`)
- [x] create new Logos and Verbum desktop icons (to steer clear of copyright issues; we currently use actual FaithLife logos)
- [x] ensure that GH Actions still build the binary correctly
- [x] ensure that the `Dockerfile` still builds the binary correctly
- [x] update documentation to use app/package/binary names


thw26 (changes requested):
All of this looks good.

It looks like we just need to resolve the other checklist bits in the OP and we should be good to do some final tests.

thw26:
As part of this, we might want to fix #171, since we need to develop our own icons.

thw26 (approved):

Issue #23, opened 2023-12-30
Migrate and Update the Wiki


thw26:
https://github.com/ferion11/LogosLinuxInstaller/wiki

We need to update the Wiki, preferably before release. This is a good time to prune the information and provide additional FAQ-type material.

PR #201, opened 2024-10-23
docs: update README with new invocation

ctrlaltf24:
There is no ./main.py anymore. Perhaps we should re-word this documentation to use the pypi package instead


thw26:
Yes! But duplicate. See #7, #23.

Issue #170, opened 2024-09-14
Simplify winetricks choices


n8marti:
It seems unlikely that winetricks would be used anywhere other than 1) downloaded by LLI to the INSTALLDIR/data/bin folder or 2) installed at the system level. Attempting to handle an arbitrary path, then, for WINETRICKSBIN could lead to more maintenance effort than is worth it.

Proposal: Only allow two options for WINETRICKSBIN: "default"/"download" or "system". Or, even simpler, leave it undefined for the default value, or accept "system" if the user wants to override that with the system version. In order to be backwards compatible, though, we would need a conversion function to convert legacy paths to either "system" or "default".

n8marti:
I've discovered that our wine appimage actually includes winetricks. So an even further simplification would be that "default" uses that one, while "custom" could be an arbitrary path given by the user. I think this choice only matters to developers/testers, so I would prefer to not even include it in the GUI, which would just automatically use the appimage winetricks.

Issue #155, opened 2024-08-03
Store Logos Releases in a Cache File


thw26:
I'm thinking in .local/state or somewhere similar. Ideally if we do this, anytime an update is run, we would compare the cache file to what is on the web, and either append the new or replace the whole. We could probably reduce the rate at which releases are checked for to once a week unless a force flag is passed.

thw26:
We could do something similar with GitHub data and the md5sums to reduce GitHub API calls.

PR #205, opened 2024-10-23
refactor: add type hints and comments

ctrlaltf24:
Small handful of comments on things I noticed while reading

No functional changes

### DIFF
--- a/ou_dedetai/config.py
+++ b/ou_dedetai/config.py
@@ -3,6 +3,7 @@ import logging
 import os
 import tempfile
 from datetime import datetime
+from typing import Optional
 
 
 # Define app name variables.
@@ -59,20 +60,20 @@ for key, default in extended_config.items():
     globals()[key] = os.getenv(key, default)
 
 # Set other run-time variables not set in the env.
-ACTION = 'app'
-APPIMAGE_FILE_PATH = None
-authenticated = False
-BADPACKAGES = None
+ACTION: str = 'app'
+APPIMAGE_FILE_PATH: Optional[str] = None
+authenticated = False # FIXME: no references
+BADPACKAGES: Optional[str] = None # FIXME: no references
 DEFAULT_CONFIG_PATH = os.path.expanduser(f"~/.config/FaithLife-Community/{name_binary}.json")  # noqa: E501
-FLPRODUCTi = None
-GUI = None
-INSTALL_STEP = 0
-INSTALL_STEPS_COUNT = 0
+FLPRODUCTi: Optional[str] = None
+GUI = None # FIXME: no references
+INSTALL_STEP: int = 0
+INSTALL_STEPS_COUNT: int = 0
 L9PACKAGES = None
 LEGACY_CONFIG_FILE = os.path.expanduser("~/.config/Logos_on_Linux/Logos_on_Linux.conf")  # noqa: E501
 LLI_AUTHOR = "Ferion11, John Goodman, T. H. Wright, N. Marti"
 LLI_CURRENT_VERSION = "4.0.0-beta.2"
-LLI_LATEST_VERSION = None
+LLI_LATEST_VERSION: Optional[str] = None
 LLI_TITLE = name_app
 LOG_LEVEL = logging.WARNING
 LOGOS_BLUE = '#0082FF'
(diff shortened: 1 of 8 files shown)


ou_dedetai/config.py
```diff
@@ -59,55 +60,55 @@
...
 
 # Set other run-time variables not set in the env.
-ACTION = 'app'
-APPIMAGE_FILE_PATH = None
-authenticated = False
-BADPACKAGES = None
+ACTION: str = 'app'
+APPIMAGE_FILE_PATH: Optional[str] = None
+authenticated = False # FIXME: no references
+BADPACKAGES: Optional[str] = None # FIXME: no references
```
thw26, line 66: This is leftover from when we were still using this for removing AppImageLauncher. We ended up having to handle AppImageLauncher differently, but we kept this in case we ever discovered a problematic package.

ou_dedetai/config.py
```diff
@@ -59,55 +60,55 @@
...
-MYDOWNLOADS = None
-OS_NAME = None
-OS_RELEASE = None
-PACKAGE_MANAGER_COMMAND_INSTALL = None
-PACKAGE_MANAGER_COMMAND_REMOVE = None
-PACKAGE_MANAGER_COMMAND_QUERY = None
-PACKAGES = None
-PASSIVE = None
+LOGOS10_RELEASES = None     # used to save downloaded releases list # FIXME: not set #noqa: E501
+MYDOWNLOADS: Optional[str] = None # FIXME: Should this use the tempfile module?
```
thw26, line 93: The main reason this exists is for reuse_download. Our goal is to minimize network bandwidth for Christian workers in low bandwidth/data limited scenarios.

ou_dedetai/config.py
```diff
@@ -59,55 +60,55 @@
     globals()[key] = os.getenv(key, default)
 
 # Set other run-time variables not set in the env.
-ACTION = 'app'
-APPIMAGE_FILE_PATH = None
-authenticated = False
-BADPACKAGES = None
+ACTION: str = 'app'
+APPIMAGE_FILE_PATH: Optional[str] = None
+authenticated = False # FIXME: no references
```
thw26, line 65: This was from a time when I had implemented password requests using the system keyring, authenticated meaning user password authenticated. At the time we chose to drop this functionality for the sake of user trust and we opted for using pkexec in most situations.

ou_dedetai/config.py
```diff
@@ -59,55 +60,55 @@
...
+BADPACKAGES: Optional[str] = None # FIXME: no references
 DEFAULT_CONFIG_PATH = os.path.expanduser(f"~/.config/FaithLife-Community/{name_binary}.json")  # noqa: E501
-FLPRODUCTi = None
-GUI = None
-INSTALL_STEP = 0
-INSTALL_STEPS_COUNT = 0
+FLPRODUCTi: Optional[str] = None
+GUI = None # FIXME: no references
```
thw26, line 69: This most likely can be safely removed. I think it would have been used to keep track of the GUI object, but we have the `app` var we pass around now.

ou_dedetai/config.py
```diff
@@ -59,55 +60,55 @@
...
-LOGOS_ICON_FILENAME = None
-LOGOS_ICON_URL = None
+LOGOS_FORCE_ROOT: bool = False
+LOGOS_ICON_FILENAME: Optional[str] = None
+LOGOS_ICON_URL: Optional[str] = None
 LOGOS_LATEST_VERSION_FILENAME = name_binary
-LOGOS_LATEST_VERSION_URL = None
-LOGOS9_RELEASES = None      # used to save downloaded releases list
+LOGOS_LATEST_VERSION_URL: Optional[str] = None
+LOGOS9_RELEASES = None      # used to save downloaded releases list # FIXME: not set #noqa: E501
```
thw26, line 89: See #155.

ou_dedetai/config.py
```diff
@@ -59,55 +60,55 @@
...
-LOGOS10_RELEASES = None     # used to save downloaded releases list
-MYDOWNLOADS = None
-OS_NAME = None
-OS_RELEASE = None
-PACKAGE_MANAGER_COMMAND_INSTALL = None
-PACKAGE_MANAGER_COMMAND_REMOVE = None
-PACKAGE_MANAGER_COMMAND_QUERY = None
-PACKAGES = None
-PASSIVE = None
+LOGOS10_RELEASES = None     # used to save downloaded releases list # FIXME: not set #noqa: E501
```
thw26, line 92: See #155.

ou_dedetai/control.py
```diff
@@ -38,6 +38,7 @@ def restore(app=None):
     backup_and_restore(mode='restore', app=app)
 
 
+# FIXME: consider moving this into it's own file/module.
```
thw26, line 41: I think utils is the place this probably should be in the current file structure.

Its existence here is a holdover from the bash script.

This functionality I think could easily be in its own module, or perhaps we could refactor things to have our various copy functions or otherwise all in a files.py or something similar.

ou_dedetai/control.py
```diff
@@ -268,8 +269,8 @@ def set_winetricks():
             # Check if local winetricks version is up-to-date; if so, offer to
             # use it or to download; else, download it.
             local_winetricks_version = subprocess.check_output(["winetricks", "--version"]).split()[0]  # noqa: E501
-            if str(local_winetricks_version) >= "20220411":
-                if config.DIALOG == 'tk':
+            if str(local_winetricks_version) >= "20220411": #FIXME: consider using config.WINETRICKS_VERSION and != string comparision on versions is dodgy #noqa: E501
+                if config.DIALOG == 'tk': #FIXME: CLI client not considered
```
thw26, line 273: See #170, which may affect this

ou_dedetai/control.py
```diff
@@ -319,7 +321,7 @@ def set_winetricks():
     return 0
 
 
-def download_winetricks():
+def download_winetricks(): #FIXME: unused, logic moved to system.install_winetricks
```
thw26, line 324: Good catch. I'm sure this was left behind due to a bad rebase.

ou_dedetai/system.py
```diff
@@ -259,7 +260,7 @@ def reboot():
     sys.exit(0)
 
 
-def t(command):
+def t(command): # FIXME: unused, have_dep does the same thing
```
thw26, line 263: I've noticed this before but didn't give it much thought. Good to replace one with another.

ou_dedetai/system.py
```diff
@@ -275,6 +276,7 @@ def tl(library):
 
 
 def get_dialog():
+    # FIXME: wouldn't the cli still work without a DISPLAY? What about wayland?
```
thw26, line 279: This is about when executing the wine functions that require a GUI, not necessarily our CLI. We'd need to make this specified based on the CLI command run.

ou_dedetai/utils.py
```diff
@@ -225,7 +225,7 @@ def delete_symlink(symlink_path):
             logging.error(f"Error removing symlink: {e}")
 
 
-def check_dependencies(app=None):
+def check_dependencies(app=None): # FIXME: misnomer, isn't this an install_dependencies
```
thw26, line 228: Holdover from a refactor

ou_dedetai/utils.py
```diff
@@ -442,7 +442,7 @@ def get_winetricks_options():
         # Check if local winetricks version is up-to-date.
         cmd = ["winetricks", "--version"]
         local_winetricks_version = subprocess.check_output(cmd).split()[0]
-        if str(local_winetricks_version) >= "20220411":
+        if str(local_winetricks_version) >= "20220411": #FIXME: consider using config.WINETRICKS_VERSION and != string comparision on versions is dodgy #noqa: E501
```
thw26, line 445: Sounds like a good improvement.

thw26 (commented):
Reviewed your comments. A lot sound like good TODOs/FIXMEs. Left my thoughts on others that might need more work.

ou_dedetai/config.py
ctrlaltf24: Ah. Perhaps a separate directory? ~/.cache/FaithLife-Community/...?

ou_dedetai/control.py
ctrlaltf24: backup.py makes sense to me, there's a lot of logic here

Issue #168, opened 2024-09-13
Reload Config Option


thw26:
We should see if it is possible to make it so our config can be reloaded at run time, in case a user manually edits the config while the program is running, e.g., to switch the release channel (cf. #157; #163)

Issue #35, opened 2024-01-05
Clean up #TODO and #FIXME


thw26:
As we code, we might add TODOs and FIXMEs along the way. Verify that none of these remain, or most of them are fixed, or outstanding ones are converted to git issues.

Issue #187, opened 2024-10-04
Split Persistent Config into User and Internal


thw26:
Right now our config stores non-user editable fields. We should have a second log file that stores system level vars in `~/.local/state` so as not to confuse users who edit the config manually.

We may want to make a function/optarg that outputs other environment vars that a user can set at the CLI, so as not to clutter the main `--help` but to be accessible in program.

PR #206, opened 2024-10-24
feat: add abstract app module and refactor config

ctrlaltf24:
# Improvements
## Type Enforcement
Types are checked using a static analysis tool (mypy) to ensure nothing is None when it shouldn't be (or a Path when it should be a string)
## Network Caching
All network requests are cached

## Config Enforcement
Wrapper struct to ensure all config variables are set, if not the user is prompted (invalid answers are prompted again).

Additionally configuration hooks were added so all UIs can refresh themselves if the config changes underneath them. A reload function has also been added (TUI only as it's for power users and offers no benefit to the cli)

## Config Format Change
In the interest of consistency of variable names, the config keys have changed. Legacy values will continue to be read (and written). New keys take precedence. This new config can be copied to older versions of the installer and it should (mostly) work (not aware of any deficiencies however downgrading is hard to support). Legacy paths are moved to the new location.

## Graphical User Interface (tinker)
Ensured that the ask function was never called. We want to use the given UI, as it's prettier. However if the case should arise where we accidentally access a variable before we set it, the user will see a pop-up with the one question. Data flow has been changed to pull all values from the config, it no longer stores copies. It also now populates all drop-downs in real-time as other drop-downs are changed, as a result there was no longer any need for the "Get EXE" and "Get Release List" buttons, so they were removed. Progress bar now considers the entire installation rather than "completing" after each step.

## Terminal User Interface (curses)
Appears the same as before from the user's standpoint, however now there is a generic ask page, greatly cutting down on the number of queues/Events. There may be some more unused Queues/Events lingering, didn't see value in cleaning them up.

## Command Line Interface
Prompts appears the same as before, progress bar is now prepended to status lines

## Misc other changes
- renamed arch p7zip package to 7zip
- use winecfg rather than winetricks to set win version

## Closing Notes
One goal of this refactor was to keep the code from being understood by someone who is already familiar with it. There is a couple new concepts, the abstract base class, reworked config, and network cache, but the core logic of how it works remains unchanged. If something isn't clear please ask.

# Screenshots
## GUI Sample
GUI-sample-prompt
## TUI Sample
TUI-prompt
## CLI Sample
CLI-prompt
## GUI behavior
Before Product is selected
GUI-before-product-selected
After product is selected
GUI-after-product-selected

Fixes: #147, #234, #35, #168, #155


ou_dedetai/tui_app.py
```diff
@@ -719,25 +713,25 @@ def switch_screen(self, dialog):
...
+    def handle_ask_response(self, choice: Optional[str]):
+        if choice is not None:
+            self.ask_answer_queue.put(choice)
+            self.ask_answer_event.set()
+            self.switch_screen(config.use_python_dialog)
 
     def get_version(self, dialog):
-        self.product_e.wait()
-        question = f"Which version of {config.FLPRODUCT} should the script install?"  # noqa: E501
+        question = f"Which version of {self.conf.faithlife_product} should the script install?"  # noqa: E501
```
thw26, line 734: To my knowledge, I need to wait on self.product_e.wait() as if not, the TUI charges through the installer process; should this way now be handled by the TUI's implementation of app._hook_product_update()?

thw26:
Thanks for this! The framework you have here looks great. The TUI is a Frankenstein of my own thought. As mentioned to Nate, I see tui_screen.py as a library that could feasibly be used outside of our project, so also the same with certain aspects of the display code in tui_app.py, say lines https://github.com/FaithLife-Community/LogosLinuxInstaller/blob/0def2e5b99225db5de97fea97acaaaa0d8efa52b/ou_dedetai/tui_app.py#L1-L350. The task processor code in tui_app and in gui_app could also be brought into the abstract class. I would also hope that the choice_processor code in tui_app.py might find its way there.

@n8marti has handled the GUI, so I will leave that to him. (Given your review, you might be the third person we've needed: someone who understands both the TUI and the GUI, haha.)

I originally tried to code for the various UIs particularly in msg.py. This eventually got away from me and found its way back in msg.status(). There's likely a fair chunk of room for messaging to be brought into the abstract class given how much code is relatively unused in that module.

I had also tried this in installer, but the GUI needed enough odds and ends to be separated at the time. Now that we have our working base, I think it'd be great to try to reel in the various odd bits and make these more united, all in the spirit of #147.

As mentioned elsewhere, this would also be helpful for #2 and #87, and for drastically improving code reusability/maintenance. Given your further comments about the suggested config changes, that would go well with #187. While I think all these issues are too much for one PR, I do think we could lump #147 and #187 into this PR's scope as a way of refactoring our code.

ou_dedetai/tui_app.py
ctrlaltf24: Surprisingly enough it didn't charge through - it waited for the question to be answered before going to the next screen probably because we have two main threads in the TUI:

1. The input processing thread (main, don't block me)
2. When events are triggered they're spawned on a new thread (this is the installation thread). Since the ask call is blocking, this gets stopped while the user is inputting their value (which the processing is done on the first thread)

Then if you notice in TUI's ask implementation there is an event wait to communicate between the two threads

n8marti:
Am I understanding correctly that in the GUI the user will see the "Choose which FaithLife product" window first, then they will see the GUI installer window, with all the options pre-populated with defaults? So then they can make adjustments, or just click "Install"?

ctrlaltf24:
> Am I understanding correctly that in the GUI the user will see the "Choose which FaithLife product" window first, then they will see the GUI installer window, with all the options pre-populated with defaults? So then they can make adjustments, or just click "Install"?

Not quite, I liked the current GUI flow so much I didn't want to modify it. Before and after this PR it behaves the same, however if someday in the future there was a code path that tried to retrieve the faithlife product before the prompt showed up, it would open a separate dialog asking that question. In the GUI's case we probably want to avoid this, however the code will handle that case and avoid an error if such a code path were to exist in the future.

n8marti:
> > Am I understanding correctly that in the GUI the user will see the "Choose which FaithLife product" window first, then they will see the GUI installer window, with all the options pre-populated with defaults? So then they can make adjustments, or just click "Install"?
> 
> Not quite, I liked the current GUI flow so much I didn't want to modify it. Before and after this PR it behaves the same, however if someday in the future there was a code path that tried to retrieve the faithlife product before the prompt showed up, it would open a separate dialog asking that question. In the GUI's case we probably want to avoid this, however the code will handle that case and avoid an error if such a code path were to exist in the future.

Okay, I'm happy with that.

n8marti:
If you @ctrlaltf24 can take care of the potential merge conflicts, then I'll look it over again for approval.

ctrlaltf24:
Expanding usage of this framework....

thw26 (commented):
Additional changes needed:

- reload config
- store network cache in a separate file (in local state)

thw26:
Additional work:

- Bring over Alpine changes (#24; cf. #183)
- Bring over Architecture detection, once merged (#119)

ctrlaltf24:
> Additional work:
> 
>     * Bring over Alpine changes ([Add ARM/AARCH64 Support #24](https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/24); cf. [Fix #17. Add Alpine Support #183](https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/183))
> 
>     * Bring over Architecture detection, once merged ([Add Architecture Detection #119](https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/119))

done

ctrlaltf24:
Ready for review @thw26 @n8marti . I've done some testing, and plan to do more. You're welcome to test, everything's expected to work.

Please do not merge until I have completed my testing

ou_dedetai/wine.py
```diff
@@ -431,62 +414,69 @@ def set_win_version(exe, windows_version):
...
+
+# FIXME: Consider when to re-run this if it changes.
+# Perhaps we should have a "apply installation updates"
+# or similar mechanism to ensure all of our latest methods are installed
+# including but not limited to: system packages, winetricks options,
+# icu files, fonts, registry edits, etc.
+#
+# Seems like we want to have a more holistic mechanism for ensuring
+# all users use the latest and greatest.
+# Sort of like an update, but for wine and all of the bits underneath "Logos" itself
```
thw26, line 435: These are fair comments, and certainly what the code has been building toward. But we would need some way of determining, what version of such and such winetricks command did we run, should we need to run it again.

Most of our prior validation of install was either determining a file or an existing Windows key.

The problem we get into with some of this has been, what if a user brings an install that we know nothing about? The config was the solution.

ou_dedetai/constants.py
```diff
@@ -0,0 +1,50 @@
...
+DEFAULT_APP_WINE_LOG_PATH= os.path.expanduser("~/.local/state/FaithLife-Community/wine.log")  # noqa: E501
+DEFAULT_APP_LOG_PATH= os.path.expanduser(f"~/.local/state/FaithLife-Community/{BINARY_NAME}.log")  # noqa: E501
+NETWORK_CACHE_PATH = os.path.expanduser("~/.cache/FaithLife-Community/network.json")  # noqa: E501
+DEFAULT_WINEDEBUG = "fixme-all,err-all"
+LEGACY_CONFIG_FILES = [
+    os.path.expanduser("~/.config/FaithLife-Community/Logos_on_Linux.json"),
+    os.path.expanduser("~/.config/Logos_on_Linux/Logos_on_Linux.json"),
+    os.path.expanduser("~/.config/Logos_on_Linux/Logos_on_Linux.conf")
+]
+LLI_AUTHOR = "Ferion11, John Goodman, T. H. Wright, N. Marti"
```
thw26, line 32: Add N. Shaaban

ou_dedetai/constants.py
```diff
@@ -0,0 +1,50 @@
...
+LLI_CURRENT_VERSION = "4.0.0-beta.4"
```
thw26, line 33: Bump to 4.0.0-beta.5

Update CHANGELOG since last release.

ou_dedetai/config.py
```diff
@@ -1,209 +1,1052 @@
...
+        if self._raw.wine_binary is not None and not Path(self._raw.wine_binary).exists() and (Path(self.install_dir) / self._raw.wine_binary).exists(): # noqa: E501
+            return str(Path(self.install_dir) / self._raw.wine_binary)
+        return output
+
+    @wine_binary.setter
+    def wine_binary(self, value: str):
+        """Takes in a path to the wine binary and stores it as relative for storage"""
+        # Make the path absolute for comparison
+        relative = self._relative_from_install_dir(value)
+        # FIXME: consider this, it doesn't work at present as the wine_binary may be an
```
thw26, line 776: Relative paths and appimage are used so as not to break the appimage symlinks in the event the installer folder is moved after install

ou_dedetai/config.py
```diff
@@ -1,209 +1,1052 @@
...
+    @property
+    def wine64_binary(self) -> str:
+        return str(Path(self.wine_binary).parent / 'wine64')
+    
+    @property
+    # This used to be called WINESERVER_EXE
+    def wineserver_binary(self) -> str:
+        return str(Path(self.wine_binary).parent / 'wineserver')
+
+    # FIXME: seems like the logic around wine appimages can be simplified
```
thw26, line 822: Possibly. These are handled differently in part because of how they are validated and we've noticed that appimage wine bottles and non-appimage wine bottles don't place nicely if you swap the binaries around.

ou_dedetai/app.py
```diff
@@ -0,0 +1,261 @@
...
+from ou_dedetai import constants
+from ou_dedetai.constants import (
+    PROMPT_ANSWER_COME_AGAIN,
+    PROMPT_OPTION_DIRECTORY,
+    PROMPT_OPTION_FILE
+)
+
+
+class App(abc.ABC):
+    # FIXME: consider weighting install steps. Different steps take different lengths
```
thw26, line 19: This would be a good improvement; alternatively, split up long steps into smaller steps.

ou_dedetai/tui_curses.py
```diff
@@ -2,60 +2,74 @@
...
-from . import msg
-from . import utils
+from ou_dedetai import tui_screen
+from ou_dedetai.app import App
 
 
-def wrap_text(app, text):
+# NOTE to reviewer: does this convay the original meaning?
+# The usages of the function seemed to have expected a list besides text_centered below
+# Which handled the string case. Is it faithful to remove the string case?
```
thw26, line 11: The main purpose of this function is to break up a line that is too long, otherwise it will cause curses to throw an error for trying to draw off screen. Eventually I implemented a catch-all try/except for these errors, but I attempted to fix them before doing so as much as possible.

ou_dedetai/tui_curses.py
```diff
@@ -2,60 +2,74 @@
...
+def write_line(app: App, stdscr: curses.window, start_y, start_x, text, char_limit, attributes=curses.A_NORMAL): #noqa: E501
+    from ou_dedetai.tui_app import TUI
+    if not isinstance(app, TUI):
+        raise ValueError("curses MUST be used with the TUI")
     try:
         stdscr.addnstr(start_y, start_x, text, char_limit, attributes)
     except curses.error:
-        signal.signal(signal.SIGWINCH, app.signal_resize)
+        # FIXME: what do we want to do to handle this error?
+        # Before we were registering a signal handler
```
thw26, line 34: This exists because of the above: if curses tries to write beyond the screen limits, it will error. The resize is meant to try to correct it by resizing the screen, say, should the screen be moved around the user's desktop, perhaps temporarily shrinking in size, or manually shrinking it down and then back up again.

I'm open to suggestions.

ou_dedetai/tui_curses.py
```diff
@@ -303,12 +301,13 @@ def input(self):
...
                         self.do_menu_down()
             elif key == ord('\n') or key == 10:  # Enter key
-                self.user_input = self.options[config.current_option]
+                self.user_input = self.options[self.app.current_option]
             elif key == ord('\x1b'):
                 signal.signal(signal.SIGINT, self.app.end)
-            else:
-                msg.status("Input unknown.", self.app)
-                pass
+            # FIXME: do we need to log this?
```
thw26, line 307: No, this is for development.

ou_dedetai/tui_app.py
```diff
@@ -23,109 +32,198 @@
...
+        self._console: Optional[tui_screen.ConsoleScreen] = None
+        # End internal property values
+
+        # Lines for the on-screen console log
+        self.console_log: list[str] = []
+
+        # Note to reviewers:
+        # This does expose some possibly untested code paths
+        #
+        # Before some function calls didn't pass use_python_dialog falling back to False
```
thw26, line 113: That was intentional, as the main menu I wanted to run in curses so that someone got the feedback which dialog couldn't give.

ou_dedetai/tui_app.py
```diff
@@ -23,109 +32,198 @@
...
+                    "The 'dialog' package was not found. Falling back to Python Curses."
+                )  # noqa: E501
+            elif result:
+                logging.debug("Dialog version is up-to-date.")
+                self.use_python_dialog = True
+            else:
+                logging.error(
+                    "Dialog version is outdated. The program will fall back to Curses."
+                )  # noqa: E501
+        # FIXME: remove this hard-coding after considering whether we want to continue
```
thw26, line 131: It's a good question. The main reason, as explained in another comment, was for TUI file/directory selection.

ou_dedetai/tui_app.py
```diff
@@ -349,107 +463,100 @@ def run(self):
...
+        self.total_pages = 0
 
     def go_to_main_menu(self):
+        self.reset_screen()
         self.menu_screen.choice = "Processing"
         self.choice_q.put("Return to Main Menu")
 
     def main_menu_select(self, choice):
+        def _install():
+            self.status("Installing...")
```
thw26, line 526: Use this char: …

Looks nicer

ou_dedetai/tui_app.py
```diff
@@ -458,202 +565,171 @@ def main_menu_select(self, choice):
...
     def winetricks_menu_select(self, choice):
         if choice == "Download or Update Winetricks":
             self.reset_screen()
-            control.set_winetricks()
+            control.set_winetricks(self)
             self.go_to_main_menu()
         elif choice == "Run Winetricks":
             self.reset_screen()
-            wine.run_winetricks()
+            self.status("Running winetricks...")
```
thw26, line 605: Same

ou_dedetai/tui_app.py
```diff
@@ -458,202 +565,171 @@ def main_menu_select(self, choice):
...
+            wine.run_winetricks(self)
             self.go_to_main_menu()
         elif choice == "Install d3dcompiler":
             self.reset_screen()
-            wine.install_d3d_compiler()
+            self.status("Installing d3dcompiler...")
```
thw26, line 610: Same. Not sure if there are others missed

ou_dedetai/tui_app.py
```diff
@@ -458,202 +565,171 @@ def main_menu_select(self, choice):
...
-            utils.update_to_latest_recommended_appimage()
+            utils.update_to_latest_recommended_appimage(self)
             self.go_to_main_menu()
         elif choice == "Set AppImage":
             # TODO: Allow specifying the AppImage File
-            appimages = utils.find_appimage_files(utils.which_release())
-            appimage_choices = [["AppImage", filename, "AppImage of Wine64"] for filename in
-                                appimages]  # noqa: E501
+            appimages = self.conf.wine_app_image_files
+            # NOTE to reviewer: is this logic correct?
```
thw26, line 698: Needs tested

ou_dedetai/tui_app.py
```diff
@@ -669,215 +745,112 @@ def password_prompt(self, choice):
...
-                self.screen_q.put(self.stack_text(13, self.todo_q, self.todo_e,
-                                                  "Installing dependencies…\n", wait=True,
-                                                  dialog=config.use_python_dialog))
-
     def renderer_select(self, choice):
         if choice in ["gdi", "gl", "vulkan"]:
             self.reset_screen()
-            wine.set_renderer(choice)
-            msg.status(f"Changed renderer to {choice}.", self)
+            self.status(f"Changing renderer to {choice}.")
```
thw26, line 751: Add the 0 as below.

thw26 (commented):
Initial review complete. I need to test.

ou_dedetai/msg.py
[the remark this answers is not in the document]
ctrlaltf24: It still has logging setup. We could renaming this to logging, but not much use

ou_dedetai/main.py
```diff
@@ -317,102 +346,46 @@ def run_control_panel():
...
-    if config.DIALOG is None:
-        system.get_dialog()
-    else:
-        config.DIALOG = config.DIALOG.lower()
+    return parse_args(cli_args, parser)
 
-    if config.DIALOG == 'curses' and "dialog" in sys.modules and config.use_python_dialog is None:  # noqa: E501
-        config.use_python_dialog = system.test_dialog_version()
 
-        if config.use_python_dialog is None:
```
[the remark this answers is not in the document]
ctrlaltf24: If we could support tab completion, we would have a need for dialog then? Seems like an easier avenue than supporting two possibilities

ou_dedetai/utils.py
```diff
@@ -225,76 +109,39 @@ def delete_symlink(symlink_path):
             logging.error(f"Error removing symlink: {e}")
 
 
-def install_dependencies(app=None):
-    if config.TARGETVERSION:
-        targetversion = int(config.TARGETVERSION)
+# FIXME: should this be in control?
```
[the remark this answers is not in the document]
ctrlaltf24: I thought of control as the main action area, with utilities in the other files (network, utils, system)

ou_dedetai/utils.py
```diff
@@ -711,15 +470,17 @@ def check_appimage(filestr):
...
     directories = [
         os.path.expanduser("~") + "/bin",
-        config.APPDIR_BINDIR,
-        config.MYDOWNLOADS
+        app.conf.installer_binary_dir,
+        app.conf.download_dir
     ]
-    if config.CUSTOMBINPATH is not None:
-        directories.append(config.CUSTOMBINPATH)
+    # FIXME: consider what we should do with this, promote to top level config?
```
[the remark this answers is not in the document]
ctrlaltf24: Not needed, IMO as-is is fine. We can remove the comment

```suggestion
```

ou_dedetai/wine.py
```diff
@@ -1,92 +1,70 @@
...
-    if None in logos_cmds:
-        config.logos_cef_cmd = f'C:\\users\\{config.wine_user}\\AppData\\Local\\Logos\\System\\LogosCEF.exe'  # noqa: E501
-        config.logos_indexer_cmd = f'C:\\users\\{config.wine_user}\\AppData\\Local\\Logos\\System\\LogosIndexer.exe'  # noqa: E501
-        config.logos_login_cmd = f'C:\\users\\{config.wine_user}\\AppData\\Local\\Logos\\System\\Logos.exe'  # noqa: E501
-    config.logos_indexer_exe = str(Path(utils.find_installed_product()).parent / 'System' / 'LogosIndexer.exe')  # noqa: E501
-
-
-def check_wineserver():
+def check_wineserver(app: App):
+    # FIXME: if the wine version changes, we may need to restart the wineserver
```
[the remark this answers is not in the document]
ctrlaltf24: I've had cases where msiexec was lingering from a previous install attempt

ou_dedetai/tui_curses.py
ctrlaltf24: pass is probably good enough for now, as we recover from the error

ou_dedetai/main.py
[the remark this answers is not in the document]
thw26: If we can, then let's do it!

ou_dedetai/wine.py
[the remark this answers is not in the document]
thw26: I have seen that, too

ou_dedetai/main.py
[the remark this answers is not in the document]
ctrlaltf24: Supported basic tab completion here. See commit message for limitations

https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/206/commits/db4f1756dbdd8977781205b439ff566013a1af23

ou_dedetai/tui_curses.py
ctrlaltf24: sounds good, removing comment

n8marti (changes requested):
All in all I'm enjoying this broad refactor. I think it will put things on a more stable footing for maintainability. Let's get it polished off!

This is only a partial review b/c of time constraints, but there are several things to discuss and change. The biggest ones are:
- no more output being written to the log file
- the Installer GUI should still include status and progress, because it's strange to find that sent to the Control Panel instead
- I think we should decide on a formatting standard and set ourselves up to follow it. We've been vacillating between pep8 and not, and it's making things inconsistent. I don't have a preference, just that we make a decision and stick with it.
- wine binary options are listed in preferred order, with the newest appimage as the default. This will cause issues.

ou_dedetai/gui_app.py
```diff
@@ -142,55 +232,78 @@ def __init__(self, new_win, root, **kwargs):
...
+        self.conf.install_dir = self.conf.install_dir_default
 
     def get_winetricks_options(self):
-        config.WINETRICKSBIN = None  # override config file b/c "Download" accounts for that  # noqa: E501
-        self.gui.tricks_dropdown['values'] = utils.get_winetricks_options()
+        # override config file b/c "Download" accounts for that
+        # Type hinting ignored due to https://github.com/python/mypy/issues/3004
+        self.conf.winetricks_binary = None # type: ignore[assignment]
+        self.gui.tricks_dropdown['values'] = utils.get_winetricks_options() #noqa: E501
         self.gui.tricksvar.set(self.gui.tricks_dropdown['values'][0])
```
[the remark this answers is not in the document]
thw26: We just need to be mindful of system binaries being used, which will need a non-appimage solution.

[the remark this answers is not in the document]
n8marti: good point. So download winetricks if the chosen wine binary is not an appimage.

ou_dedetai/gui.py
```diff
@@ -69,59 +89,47 @@ def __init__(self, root, **kwargs):
...
         self.okay_button = Button(self, text="Install")
         self.okay_button.state(['disabled'])
 
-        # Status area.
-        s1 = Separator(self, orient='horizontal')
-        self.statusvar = StringVar()
-        self.status_label = Label(self, textvariable=self.statusvar)
-        self.progressvar = IntVar()
-        self.progress = Progressbar(self, variable=self.progressvar)
-
```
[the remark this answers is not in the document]
ctrlaltf24: We have a couple options:
- Close the installer window after it started the install (and gray out the install button), showing the progress on the control window. This may be useful if we ever have an "express install button" where the user doesn't select any preferences, it just goes with the default
- Make it look how it did before, using a conditional to figure out whether or not to send status messages to the installer window or control window

ctrlaltf24:
> I think we should decide on a formatting standard and set ourselves up to follow it. We've been vacillating between pep8 and not, and it's making things inconsistent. I don't have a preference, just that we make a decision and stick with it.

I've always liked picking an auto-formatter and just letting it do it's thing rather than trying to have humans being consistent with a styling. How about https://docs.astral.sh/ruff/ ?

ctrlaltf24:
> no more output being written to the log file

fixed

>     * the Installer GUI should still include status and progress, because it's strange to find that sent to the Control Panel instead

See comment https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/206#discussion_r1873925691

> wine binary options are listed in preferred order, with the newest appimage as the default. This will cause issues.

Fixed, good catch

ou_dedetai/main.py
```diff
@@ -317,102 +346,46 @@ def run_control_panel():
...
-    if not utils.file_exists(config.CONFIG_FILE):  # noqa: E501
-        for legacy_config in config.LEGACY_CONFIG_FILES:
-            if utils.file_exists(legacy_config):
-                config.set_config_env(legacy_config)
-                utils.write_config(config.CONFIG_FILE)
-                os.remove(legacy_config)
-                break
-    else:
-        config.set_config_env(config.CONFIG_FILE)
+    msg.initialize_logging(log_level, app_log_path)
```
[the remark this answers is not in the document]
ctrlaltf24: fixed

ou_dedetai/installer.py
```diff
@@ -4,771 +4,384 @@
...
+    app.status("Creating launcher shortcuts…")
     runmode = system.get_runmode()
     if runmode == 'binary':
-        update_install_feedback("Creating launcher shortcuts…", app=app)
-        create_launcher_shortcuts()
+        app.status("Creating launcher shortcuts…")
+        create_launcher_shortcuts(app)
     else:
-        update_install_feedback(
+        # Speculation: Is this because it's hard to find the python binary?
```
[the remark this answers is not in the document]
ctrlaltf24: Ahhh updated comment to match ^

n8marti:
I like the idea of the Installer window actually just being an Installer
_configuration_ window, whose only job is to let the user select
non-default config. Then all the statusing and progressing could just
happen in the Control Panel window. But in that case I'd suggest a separate
button in the CP called "Options" or "Config" somewhere near the "Install"
button. It would disappear once the app is installed. That's where my
thinking leads to, anyway.

ou_dedetai/gui.py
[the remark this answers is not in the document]
thw26: Probably should be inversed: install or advanced install.

thw26:
> I like the idea of the Installer window actually just being an Installer _configuration_ window, whose only job is to let the user select non-default config. Then all the statusing and progressing could just happen in the Control Panel window. But in that case I'd suggest a separate button in the CP called "Options" or "Config" somewhere near the "Install" button. It would disappear once the app is installed. That's where my thinking leads to, anyway.

I responded to Nathan's comment above and to make sure it isn't lost between different locations, if we want the GUI to have an install button and a second button, perhaps it should be Install (no options, just defaults) and Advanced Install, but it sounds like six of one and half a dozen of another with your suggestion, Nate.

thw26:
> > I think we should decide on a formatting standard and set ourselves up to follow it. We've been vacillating between pep8 and not, and it's making things inconsistent. I don't have a preference, just that we make a decision and stick with it.
> 
> I've always liked picking an auto-formatter and just letting it do it's thing rather than trying to have humans being consistent with a styling. How about https://docs.astral.sh/ruff/ ?

I'm all for automation.

ou_dedetai/gui_app.py
```diff
@@ -142,55 +232,78 @@ def __init__(self, new_win, root, **kwargs):
...
+            self.conf.faithlife_product = self.gui.product_dropdown['values'][0]
+        self.gui.productvar.set(self.conf.faithlife_product)
+        if self.conf._raw.faithlife_product_version is None:
+            self.conf.faithlife_product_version = self.gui.version_dropdown['values'][-1] #noqa :E501
+        self.gui.versionvar.set(self.conf.faithlife_product_version) # noqa: E501
+
+        # Now that we know product and version are set we can download the releases
+        # And use the first one
+        # FIXME: consider what to do if network is slow, we may want to do this on a
+        # Separate thread to not hang the UI
```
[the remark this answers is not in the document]
ctrlaltf24: how slow are we talking for your internet @n8marti ?

Doing this on another thread introduces a case (that was completely removed by this refactor) where the install isn't ready immediately. rn if you load this dialog you can hit install at once, no branches, no installs not ready yet, etc. I'm not sure the delay is worth loosing that

ou_dedetai/gui.py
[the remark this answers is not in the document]
ctrlaltf24: Implemented advanced install button

n8marti:
On a good day: 1Mbps down, 500Kbps up
On a bad day: 128Kbps down, 64Kbps up
Either way, though, we have very high latency. Usually 1-4 seconds for a
successful ping to a public IP.

In my test yesterday the GUI froze for at least 5, maybe 10 seconds waiting
for the network activity to finish.

ctrlaltf24:
> On a good day: 1Mbps down, 500Kbps up On a bad day: 128Kbps down, 64Kbps up Either way, though, we have very high latency. Usually 1-4 seconds for a successful ping to a public IP. In my test yesterday the GUI froze for at least 5, maybe 10 seconds waiting for the network activity to finish.

Hmm the network cache should make this problem less bad. 

I could spawn a thread as soon as the GUI launches that populates all these defaults, so the user wouldn't be aware of the latency. Would that be an acceptable solution?

Trying not to allow the case where the installer window is shown and some values may not be initialized yet - would rather not have the branching

n8marti:
All we're really needing is to get the Logos/Verbum release list, right?
Why not show the CP window, but then deactivate the Install button until
the network activity is done?

ctrlaltf24:
> All we're really needing is to get the Logos/Verbum release list, right? Why not show the CP window, but then deactivate the Install button until the network activity is done?

My hesitation is I'd rather not end up with a case where the user gets App.ask'ed a question. If faithlife_produce_version isn't set early, it will be easier to accidentally introduce code paths that would individually prompt the user for this value. That would happen if anywhere in the app the faithlife_produce_version is accessed before it's set.

By blocking we're assured by the time the UI displays this value is set. Otherwise if it's not blocked we'd have to think about the case where it isn't set yet in at least part of the GUI.

n8marti:
It's a nice improvement in the GUI with the [Install] [Advanced install] options. Not sure about the look of it, but I don' t have any better ideas at this point.

I just did a completely clean test run: no config, no network cache file, no existing LogosBible10 folder. It failed with this result:
[oudedetai_clean-failed.log](https://github.com/user-attachments/files/18049264/oudedetai_clean-failed.log)
```
[...]
2024-12-07 17:50:59 DEBUG: wine_binary_options=['/home/nate/Téléchargements/wine-devel_9.19-x86_64.AppImage', '/home/nate/Téléchargements/wine-staging_8.14-x86_64.AppImage', '/home/nate/Téléchargements/wine-devel_8.19-x86_64.AppImage', '/home/nate/Téléchargements/wine-devel_9.12-x86_64.AppImage', '/home/nate/LogosBible10/data/bin/wine-devel_9.19-x86_64.AppImage']
2024-12-07 17:50:59 INFO: Writing config to /home/nate/.config/FaithLife-Community/oudedetai.json
2024-12-07 17:50:59 DEBUG: Installing…: None
2024-12-07 17:50:59 DEBUG: Install 7: Asking questions if needed…
2024-12-07 17:50:59 DEBUG: LLI self-update check: output=<VersionComparison.DEVELOPMENT: 3>
2024-12-07 17:50:59 DEBUG: > app.conf.faithlife_product='Logos'
2024-12-07 17:50:59 DEBUG: > app.conf.faithlife_product_version='10'
2024-12-07 17:50:59 DEBUG: > app.conf.faithlife_product_release='38.0.0.0578'
2024-12-07 17:50:59 DEBUG: > app.conf.install_dir='/home/nate/LogosBible10'
2024-12-07 17:50:59 DEBUG: > app.conf.installer_binary_dir='/home/nate/LogosBible10/data/bin'
2024-12-07 17:50:59 DEBUG: > app.conf.wine_appimage_path=PosixPath('/home/nate/Téléchargements/wine-devel_9.19-x86_64.AppImage')
2024-12-07 17:50:59 DEBUG: > app.conf.wine_appimage_recommended_url='https://github.com/FaithLife-Community/wine-appimages/releases/download/9.19-devel/wine-devel_9.19-x86_64.AppImage'
2024-12-07 17:50:59 DEBUG: > app.conf.wine_appimage_recommended_file_name='wine-devel_9.19-x86_64.AppImage'
[...]
2024-12-07 17:51:08 DEBUG: subprocess cmd: '/home/nate/Téléchargements/wine64 wineboot --init'
2024-12-07 17:51:08 ERROR: An unexpected error occurred when running ['/home/nate/Téléchargements/wine64', 'wineboot', '--init']: [Errno 2] No such file or directory: '/home/nate/Téléchargements/wine64'
```
And here's the Exception thrown in the terminal:
```
Exception in thread Ou Dedetai <function ControlWindow.run_install.<locals>._install at 0x7f51e0a66e80>:
  File "/usr/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.12/threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File "/home/nate/g/LogosLinuxInstaller-refactor/ou_dedetai/gui_app.py", line 462, in _install
    installer.install(self)
  File "/home/nate/g/LogosLinuxInstaller-refactor/ou_dedetai/installer.py", line 344, in install
    ensure_launcher_shortcuts(app)
  File "/home/nate/g/LogosLinuxInstaller-refactor/ou_dedetai/installer.py", line 328, in ensure_launcher_shortcuts
    ensure_launcher_executable(app=app)
  File "/home/nate/g/LogosLinuxInstaller-refactor/ou_dedetai/installer.py", line 306, in ensure_launcher_executable
    ensure_config_file(app=app)
  File "/home/nate/g/LogosLinuxInstaller-refactor/ou_dedetai/installer.py", line 297, in ensure_config_file
    ensure_product_installed(app=app)
  File "/home/nate/g/LogosLinuxInstaller-refactor/ou_dedetai/installer.py", line 279, in ensure_product_installed
    ensure_icu_data_files(app=app)
  File "/home/nate/g/LogosLinuxInstaller-refactor/ou_dedetai/installer.py", line 267, in ensure_icu_data_files
    ensure_winetricks_applied(app=app)
  File "/home/nate/g/LogosLinuxInstaller-refactor/ou_dedetai/installer.py", line 214, in ensure_winetricks_applied
    ensure_wineprefix_init(app=app)
  File "/home/nate/g/LogosLinuxInstaller-refactor/ou_dedetai/installer.py", line 205, in ensure_wineprefix_init
    system.wait_pid(process)
  File "/home/nate/g/LogosLinuxInstaller-refactor/ou_dedetai/system.py", line 910, in wait_pid
    os.waitpid(-process.pid, 0)
                ^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'pid'
```
I'm not sure which is the cause and which are symptoms, but it seems the "auto-config" set the wine binary as the appimage in my Downloads folder (rather than one to be copied into LogosBible10), the appimage was never copied into LogosBible10, but the symlinks were still created in LogosBible10 as if the appimage *were* there, but then the installer looked for the symlinks in the Downloads folder, which were not there:
```
$ ll ~/Téléchargements | grep wine
-rw-rw-r--  1 nate nate  595593666 Apr  7  2024 wine64_bottle.tar.gz
drwxr-xr-x  5 nate nate       4096 Dec  6 15:54 wine_9.19/
-rw-rw-r--  1 nate nate   68563244 Dec  6 16:09 wine_9.19+20.04.tar.xz
-rwxr-xr-x  1 nate nate 1138655920 Jul  6 19:43 wine-devel_8.19-x86_64.AppImage*
-rwxrw-r--  1 nate nate  211538264 Dec  6 09:21 wine-devel_9.12-x86_64.AppImage*
-rwxr-xr-x  1 nate nate  211538264 Oct  8 17:53 wine-devel_9.19-x86_64.AppImage*
-rwxrwxr-x  1 nate nate  209846616 Jun 29 05:57 wine-stable_9.0-x86_64.AppImage*
-rwxr-xr-x  1 nate nate  923726168 Aug 30  2023 wine-staging_8.14-x86_64.AppImage*
-rw-rw-r--  1 nate nate     689866 Sep 16 10:48 winetricks-20240105.zip
$ ll LogosBible10/data/bin/
total 892
drwxrwxr-x 2 nate nate   4096 Dec  7 17:51 ./
drwxrwxr-x 4 nate nate   4096 Dec  7 17:51 ../
lrwxrwxrwx 1 nate nate     33 Dec  7 17:51 selected_wine.AppImage -> ./wine-devel_9.19-x86_64.AppImage
lrwxrwxrwx 1 nate nate     24 Dec  7 17:51 wine -> ./selected_wine.AppImage
lrwxrwxrwx 1 nate nate     24 Dec  7 17:51 wine64 -> ./selected_wine.AppImage
lrwxrwxrwx 1 nate nate     24 Dec  7 17:51 wineserver -> ./selected_wine.AppImage
-rwxr-xr-x 1 nate nate 904682 Dec  7 17:51 winetricks*
```

n8marti:
Another issue we need to figure out is orphaned mounted appimages. I was suspicious, and I found dozens of them in `/tmp`. I removed them all before the clean install, and afterwards there were 4 new ones there:
```
$ mount | grep wine  
wine-devel_9.19-x86_64.AppImage on /tmp/.mount_wine64mfgdHl type fuse.wine-devel_9.19-x86_64.AppImage (ro,nosuid,nodev,relatime,user_id=1000,group_id=1001)  
wine-devel_9.19-x86_64.AppImage on /tmp/.mount_wine64cJaMiA type fuse.wine-devel_9.19-x86_64.AppImage (ro,nosuid,nodev,relatime,user_id=1000,group_id=1001)  
wine-devel_9.19-x86_64.AppImage on /tmp/.mount_winesecEKcFO type fuse.wine-devel_9.19-x86_64.AppImage (ro,nosuid,nodev,relatime,user_id=1000,group_id=1001)  
wine-devel_9.19-x86_64.AppImage on /tmp/.mount_wine-dJOcobb type fuse.wine-devel_9.19-x86_64.AppImage (ro,nosuid,nodev,relatime,user_id=1000,group_id=1001)
```
This could very well pre-date the refactor, but it still needs to be sorted out. I'll more into it tonight if I have time.

n8marti:
One minor change: There's a typo in this status message: "Setting Font Smooting to RGB..." should say "Smoothing"

ctrlaltf24:
> It's a nice improvement in the GUI with the [Install] [Advanced install] options. Not sure about the look of it, but I don' t have any better ideas at this point.
> 
> I just did a completely clean test run: no config, no network cache file, no existing LogosBible10 folder. It failed with this result: [oudedetai_clean-failed.log](https://github.com/user-attachments/files/18049264/oudedetai_clean-failed.log)

Should be fixed, try again

> This could very well pre-date the refactor, but it still needs to be sorted out. I'll more into it tonight if I have time.

likely independent, not addressing in this PR

ou_dedetai/config.py
```diff
@@ -1,209 +1,1098 @@
...
+            new_channel = "stable"
+        self._raw.app_release_channel = new_channel
+        self._write()
+    
+    @property
+    def backup_dir(self) -> Path:
+        question = "New or existing folder to store backups in: "
+        options = [PROMPT_OPTION_DIRECTORY]
+        output = Path(self._ask_if_not_found("backup_dir", question, options))
+        output.mkdir(parents=True)
```
n8marti, line 981: Need to add `exist_ok=True` to `output.mkdir`, otherwise "restore" fails b/c folder already exists.

n8marti:
Missing confirmation question during "restore". Should say something like "Restore backup from [latest folder path]?"
```
2025-01-13 18:00:07 INFO: Found 2 backup folders.
2025-01-13 18:00:07 INFO: Latest folder: /media/nate/Storage-250/LLI-data-backups/Logos10-20250113T175000

: Yes [default], No, Exit:
```

ou_dedetai/control.py
```diff
@@ -8,222 +8,167 @@
...
         m = f"Restoring backup from {str(source_dir_base)}…"
     else:
         m = f"Backing up to {str(dst_dir)}…"
-    msg.status(m, app=app)
-    msg.status("Calculating destination directory size", app=app)
-    dst_dir_size = utils.get_path_size(dst_dir)
-    msg.status("Starting backup…", app=app)
-    t = utils.start_thread(copy_data, src_dirs, dst_dir)
+    app.status(m)
+    app.status("Starting backup…")
```
n8marti, line 136: The 2nd status message should be `app.status(f"Starting {mode}…")`

n8marti: Oh, I see now that this is just a copy over from the previous code. Still seems redundant, though.

n8marti:
With a finished install and restored backup files if I click on "Run Logos" I'm not seeing the Logos window open. Here's what I have in System Monitor, several minutes after clicking the button:
image
I'll try again without using the `-P` flag to make sure Logos installs correctly, but all the files seem to be there after a quick visual check.

ctrlaltf24:
> Further GUI testing on a clean install. I ran it once successfully, but then the 2nd time the "Install" button never became activated. Here are the two log files. Maybe there's a race condition for the network check? [oudedetai_1st-run.log](https://github.com/user-attachments/files/18399579/oudedetai_1st-run.log) [oudedetai_2nd-run.log](https://github.com/user-attachments/files/18399582/oudedetai_2nd-run.log)

Fixed in https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/206/commits/9bac0f11f6f4a361b82e529d44a5edb2a68cfbfd

ou_dedetai/control.py
ctrlaltf24: Fixed in https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/206/commits/3687c778e078ed4a8534594fc06f9f43c0700f57

ou_dedetai/config.py
ctrlaltf24: done in https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/206/commits/4f725b6ad0ac7d2647bd05d17fce75887f8089ad

ctrlaltf24:
> With a finished install and restored backup files if I click on "Run Logos" I'm not seeing the Logos window open. Here's what I have in System Monitor, several minutes after clicking the button:

what does wine.log show?

ctrlaltf24:
> Missing confirmation question during "restore". Should say something like "Restore backup from [latest folder path]?"
> 
> ```
> 2025-01-13 18:00:07 INFO: Found 2 backup folders.
> 2025-01-13 18:00:07 INFO: Latest folder: /media/nate/Storage-250/LLI-data-backups/Logos10-20250113T175000
> 
> : Yes [default], No, Exit:
> ```

fixed in https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/206/commits/5e14603e6579141e5d9de3148b8670a91c967441

n8marti:
> > With a finished install and restored backup files if I click on "Run Logos" I'm not seeing the Logos window open. Here's what I have in System Monitor, several minutes after clicking the button:
> 
> what were the results of looking at the wine.log?

Nothing in wine.log besides the subprocess command. But here's what I see in `LogosBible10/data/wine64_bottle/drive_c/users/nate/AppData/Local/Faithlife/Logs/Logos/LogosError.log` after a failed run with `--run-installed-app`:
```
Program Version: 38.1 (38.1.0.0002)
Time: 2025-01-14 14:15:50 +01:00 (2025-01-14T13:15:50Z)
SQLite Error 283: recovered 10 frames from WAL file C:\users\nate\AppData\Local\Logos\Documents\dsck2rmb.rvb\LocalUserPreferences\PreferencesManager.db-wal

Program Version: 38.1 (38.1.0.0002)
Time: 2025-01-14 14:15:50 +01:00 (2025-01-14T13:15:50Z)
SQLite Error 283: recovered 10 frames from WAL file C:\users\nate\AppData\Local\Logos\Data\dsck2rmb.rvb\Errors\Application\ErrorReportManager.db-wal

Program Version: 38.1 (38.1.0.0002)
Time: 2025-01-14 14:15:50 +01:00 (2025-01-14T13:15:50Z)
SQLite Error 2058: os_win.c:49869: (33) winUnlockReadLock(C:\users\nate\AppData\Local\Logos\Data\dsck2rmb.rvb\Errors\Application\ErrorReportManager.db) - Violation de verrou.

*** Application Crash ***
Program Version: 38.1 (38.1.0.0002)
Windows Version: 10.0.19043.0
.NET Framework Version: 8.0.10
Time: 2025-01-14 14:15:50 +01:00 (2025-01-14T13:15:50Z)
Installed memory: 15,697 MB
Install path: C:\users\nate\AppData\Local\Logos\System\Logos.dll
Free install space: 0 MB
Data path: C:\users\nate\AppData\Local\Logos\Data\dsck2rmb.rvb
Free disk space: 0 MB
Temp path: C:\users\nate\AppData\Local\Temp\
Free temp space: 0 MB

Error ID: 8409
Error detail: AggregateException: One or more errors occurred. (Could not load file or assembly 'Logos.BibleStudyBuilderApi.v1, Version=3.6.0.0, Culture=neutral, PublicKeyToken=null'. Fichier introuvable.)

System.AggregateException: One or more errors occurred. (Could not load file or assembly 'Logos.BibleStudyBuilderApi.v1, Version=3.6.0.0, Culture=neutral, PublicKeyToken=null'. Fichier introuvable.)
 ---> System.IO.FileNotFoundException: Could not load file or assembly 'Logos.BibleStudyBuilderApi.v1, Version=3.6.0.0, Culture=neutral, PublicKeyToken=null'. Fichier introuvable.
File name: 'Logos.BibleStudyBuilderApi.v1, Version=3.6.0.0, Culture=neutral, PublicKeyToken=null'
   at LDLS4.Startup.StartupManager.Steps.CreateServiceClients(AsyncMethodContext context, LicenseManager licenseManager, ResourceManager resourceManager, LibraryCatalog libraryCatalog, Boolean shouldWorkOnline, User user, ApplicationFamily family, WebServiceRequestSettings requestSettings, Func`2 createFacilityClientSettings)
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at LDLS4.Startup.StartupManager.Steps.CreateServiceClients(AsyncMethodContext context, LicenseManager licenseManager, ResourceManager resourceManager, LibraryCatalog libraryCatalog, Boolean shouldWorkOnline, User user, ApplicationFamily family, WebServiceRequestSettings requestSettings, Func`2 createFacilityClientSettings)
   at LDLS4.Startup.StartupManager.StartServices(AsyncMethodContext context, User user, DigitalLibraryServices services, Boolean shouldWorkOnline, FeedbackLevel effectiveFeedbackLevel, ApplicationFamily family)
   at LDLS4.Startup.StartupManager.DoRun(AsyncMethodContext context)
   at LDLS4.Startup.StartupManager.Run(AsyncMethodContext context)
   at LDLS4.AppModel.<>c__DisplayClass691_0.<<Run>b__0>d.MoveNext()
   --- End of inner exception stack trace ---
```
Running directly from the terminal opens correctly and does not show any "Could not load" errors. However, in this same installation subsequent attempts to open with the `--run-installed-app` option seem to consistently throw this `System.OutOfMemoryException`:
```
Program Version: 38.1 (38.1.0.0002)
Time: 2025-01-14 15:04:03 +01:00 (2025-01-14T14:04:03Z)
SQLite Error 2058: os_win.c:49869: (33) winUnlockReadLock(C:\users\nate\AppData\Local\Logos\Data\dsck2rmb.rvb\Errors\Application\ErrorReportManager.db) - Violation de verrou.

*** Application Crash ***
Program Version: 38.1 (38.1.0.0002)
Windows Version: 10.0.19043.0
.NET Framework Version: 8.0.10
Time: 2025-01-14 15:04:03 +01:00 (2025-01-14T14:04:03Z)
Installed memory: 15,697 MB
Install path: C:\users\nate\AppData\Local\Logos\System\Logos.dll
Free install space: 46,158 MB
Data path: C:\users\nate\AppData\Local\Logos\Data\dsck2rmb.rvb
Free disk space: 46,158 MB
Temp path: C:\users\nate\AppData\Local\Temp\
Free temp space: 46,158 MB

Error ID: 3650
Error detail: AggregateException: One or more errors occurred. (Exception of type 'System.OutOfMemoryException' was thrown.)

System.AggregateException: One or more errors occurred. (Exception of type 'System.OutOfMemoryException' was thrown.)
 ---> System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at System.Threading.Thread.StartCore()
   at Logos.Sync.Client.SyncManager.Start()
   at LDLS4.AppModel.AppStartupSteps.StartPostSetFactoryServicesWorkCore(AsyncMethodContext context)
   at LDLS4.Startup.ExternalStartupSteps.StartPostSetFactoryServicesWork(AsyncMethodContext context)
   at LDLS4.Startup.StartupManager.StartServices(AsyncMethodContext context, User user, DigitalLibraryServices services, Boolean shouldWorkOnline, FeedbackLevel effectiveFeedbackLevel, ApplicationFamily family)
   at LDLS4.Startup.StartupManager.DoRun(AsyncMethodContext context)
   at LDLS4.Startup.StartupManager.Run(AsyncMethodContext context)
   at LDLS4.AppModel.<>c__DisplayClass691_0.<<Run>b__0>d.MoveNext()
   --- End of inner exception stack trace ---
```
I have no idea what memory that would be referring to. There are no low-memory errors in `journalctl -b`, and I hardly see the total memory usage move at all when observing System Monitor while using `--run-installed-app`.

ou_dedetai/msg.py
```diff
@@ -77,28 +74,33 @@ def initialize_logging(stderr_log_level):
...
+    app_log_path = constants.DEFAULT_APP_LOG_PATH
+    # Ensure the application log's directory exists
+    os.makedirs(os.path.dirname(app_log_path), exist_ok=True)
+
     # Ensure log file parent folders exist.
-    log_parent = Path(config.LOGOS_LOG).parent
-    if not log_parent.is_dir():
-        log_parent.mkdir(parents=True)
+    log_parent = Path(app_log_path).parent
+    log_parent.mkdir(parents=True, exist_ok=True)
```
n8marti, line 83: Isn't running `os.makedirs` redundant with `log_parent.mkdir`?

ou_dedetai/utils.py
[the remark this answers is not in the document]
n8marti: I vote for this being moved to `system.py` since it involves manipulating the OS and wouldn't be reused in any other module. In a perfect world any function here in `utils.py` would be generic and potentially used in multiple other modules.

n8marti (approved):

n8marti:
I'm very happy with the simplifications and strong foundation that this refactor brings in. I see some typos and other minor fixes that should be done, as well as a full conversion to ruff linting, but those shouldn't hold up this merge. We can do a cleanup PR in the near future.

ctrlaltf24:
Tested clean installs on TUI, GUI, and CLI one last time, all get to logos login page (needed a couple very simple bug fixes)

PR ready to Squash and merge @thw26 (using this PR's description as the commit body). Don't want these individual commits on main, as the code isn't stable in between all the commits, no reason to have the history for what in essence is a scratch pad. Commit history will remain in this PR if we need it later.

Issue #191, opened 2024-10-06
Logos 10 crashes after opening biblical word study for "worship" 


ctrlaltf24:
Reproduction steps:
- Launch Logos 10 (v36.1.0.6) with wine-devel 9.12 (or 9.16)
- Open a biblical word study for "worship" (some words cause this crash, others don't). 
- Expand the Hebrew words section
- Application crashes with the error below

```
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"Application: Logos.exe\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"CoreCLR Version: 8.0.724.31311\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L".NET Version: 8.0.7\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"Description: The application requested process termination through System.Environment.FailFast.\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"Message: Unhandled exception in async work\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"Description: The process was terminated due to an unhandled exception.System.Globalization.CultureNotFoundException: Culture is not supported. (Parameter 'name')\r\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"en-us-posix is an invalid culture identifier.\r\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"   at System.Globalization.CultureInfo.GetCultureInfo(String name)\r\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"   at System.Globalization.CultureData.IcuEnumCultures(CultureTypes types)\r\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"   at Libronix.Utility.CultureInfoUtility.CreateSpecificCulture(LanguageName language)\r\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"   at Libronix.Globalization.Culture.CreateSpecificCulture()\r\n"
2024-10-05 19:12:07 INFO: 03d0:err:eventlog:ReportEventW L"   at Libronix.Globalization.Culture.get_FormatCultureInfo()\r\n"
[...]

PR #207, opened 2024-10-24
fix: update out of date icu files

ctrlaltf24:
This is the last part needed to fix https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/191

# How to apply the update:
## TUI
Utilities -> Install ICU files
## CLI
Use flag `--install-icu`
## GUI
- App actions -> Install/update ICU files
- Run Action

### DIFF
--- a/ou_dedetai/wine.py
+++ b/ou_dedetai/wine.py
@@ -441,8 +441,9 @@ def enforce_icu_data_files(app=None):
     icu_latest_version = network.get_tag_name(json_data).lstrip('v')
 
     # This file with the tag name of the downloaded release. If it doesn't match latest override the ICU files then write this file
-    icu_version_path = Path(f"{config.WINEPREFIX}/drive_c/windows/globalization/ICU/{repo.replace('/','_')}_Version.txt")
-    if icu_version_path.exists() and icu_version_path.read_text().strip() == icu_latest_version:
+    config_dir = os.path.expanduser(f"~/.local/state/FaithLife-Community/")
+    icu_version_path = Path(f"{config_dir}/ICU_Version.txt")
+    if icu_version_path.exists() and icu_version_path.read_text().strip() == f"{repo}\n{icu_latest_version}":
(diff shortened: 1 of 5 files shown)


ou_dedetai/wine.py
```diff
@@ -434,11 +434,21 @@ def set_win_version(exe, windows_version):
...
-def install_icu_data_files(app=None):
+def enforce_icu_data_files(app=None):
     repo = "FaithLife-Community/icu"
     json_data = network.get_latest_release_data(repo)
     icu_url = network.get_first_asset_url(json_data)
-    # icu_tag_name = utils.get_latest_release_version_tag_name(json_data)
+    icu_latest_version = network.get_tag_name(json_data).lstrip('v')
+
+    # This file with the tag name of the downloaded release. If it doesn't match latest override the ICU files then write this file
+    icu_version_path = Path(f"{config.WINEPREFIX}/drive_c/windows/globalization/ICU/{repo.replace('/','_')}_Version.txt")
```
thw26, line 444: Should we store this in `"~/.local/state/FaithLife-Community/` instead?

ou_dedetai/wine.py
```diff
@@ -434,11 +434,21 @@ def set_win_version(exe, windows_version):
         wait_pid(process)
 
 
-def install_icu_data_files(app=None):
+def enforce_icu_data_files(app=None):
```
thw26, line 437: Any reason for the name change?

ctrlaltf24: Yup, now it's called all the time rather than conditionally.

It'll break early if it's already installed, hence the install name is no longer strictly accurate

ou_dedetai/wine.py
ctrlaltf24: Seems reasonable to me, I'll make the change shortly

ctrlaltf24: Fixed

PR #212, opened 2024-10-24
refactor: handfull of fixmes resolved trivally

ctrlaltf24:
Resolving some of the comments from here https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/205

### DIFF
--- a/ou_dedetai/installer.py
+++ b/ou_dedetai/installer.py
@@ -677,7 +677,15 @@ def ensure_launcher_shortcuts(app=None):
     ensure_launcher_executable(app=app)
     config.INSTALL_STEP += 1
     update_install_feedback("Creating launcher shortcuts…", app=app)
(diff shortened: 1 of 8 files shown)


ou_dedetai/installer.py
```diff
@@ -676,16 +676,8 @@ def ensure_launcher_shortcuts(app=None):
     config.INSTALL_STEPS_COUNT += 1
     ensure_launcher_executable(app=app)
     config.INSTALL_STEP += 1
-    runmode = system.get_runmode()
```
thw26, line 679: We need to verify an install when run from source that its launcher is created correctly. I believe this was set up as a workaround.

ctrlaltf24: Reverted the change, doesn't really matter

PR #215, opened 2024-10-25
docs: move complex install instructions out of README

ctrlaltf24:
As a user I didn't realize there was a simple way without git cloning when I first saw the project

Make the easy way easier to find.

The first commit has no edits, just a move, the second commit has the edit


thw26 (approved):
This seems like a great improvement to something I've tried to figure out, too

PR #219, opened 2024-10-26
feat: add *BSD support

### LINKED ISSUE #14: Add *BSD Support
The Python program is now *BSD compatible (https://github.com/ferion11/LogosLinuxInstaller/issues/150), more or less. There may still be some Linux code, but the Python app puts us in a much better location.

We need to determine what package names are for *BSD and have tests done to see if Logos can be booted on *BSD.

thw26:
Steps:

- Download GhostBSD XFCE
- Install to Disk/VM
- Boot into it
- Open terminal
- sudo pkg install -g "GhostBSD*-dev"
- sudo pkg install vim rsync git wget curl gsed gnugrep gawk samba416 7-zip cabextract wine-devel winetricks
- git clone https://github.com/FaithLife-Community/LogosLinuxInstaller
- cd LogosLinuxInstaller
- chmod 700 ./scripts/ensure*
- ./scripts/ensure-python.sh
- ensure-venv fails if run from fish and without the export
- bash
- export LD_LIBRARY_PATH=/opt/lib
- ./scripts/ensure-venv.sh
- /opt/bin/python3 -m venv venv
- source ./venv/bin/activate
- pip install .
- pip install --upgrade pip
- pip install wheel

And now you're ready to go! At least you would be, if sqlite3 was being properly loaded. This seems to be a new problem since I first worked on this and is some issue with the python ensure. I tried the following to no avail.

---
Fixes #14. There is one blocking issue and one additional problem.

I have added install pointers to the wiki. It presently requires use of the `ports` tool to get different versions of Wine and Python, but, as noted below, we don't have the Python version we need.

- ~~[x] This is blocked until FreeBSD has a working Python 3.12 package. As it stands, it is not yet ported. https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=267515 https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271673~~
- [ ] No AppImage support without: See: https://github.com/AppImage/AppImageKit/issues/98. Better solution is #316 

This is untested. This PR has been opened as a base working model. Of the testing I did do, I worked on GhostBSD.

Some of the commands we use are GNU versions of commands (e.g., sed, grep) and therefore they may not work out of the box on *BSD.

Until we can build it in GitHub, you will have to run from source or pip. Once you activate the Python virtual environment (see [CONTRIBUTING](https://github.com/FaithLife-Community/LogosLinuxInstaller/blob/main/CONTRIBUTING.md)), you will then need to install some Python modules.

For newer Wine, you may need to use ports.

### DIFF
--- a/ou_dedetai/system.py
+++ b/ou_dedetai/system.py
@@ -520,6 +520,18 @@ def get_package_manager() -> PackageManager | None:
             )
         logos_9_packages = ""
         incompatible_packages = ""  # appimagelauncher handled separately
+    elif shutil.which('pkg') is not None:  # freebsd ghostbsd
+        install_command = ["pkg", "-y", "install"]  # noqa: E501
+        download_command = ["pkg", "fetch", "-d"]  # noqa: E501
+        remove_command = ["pkg", "-y", "delete"]  # noqa: E501
+        query_command = ["pkg", "info"]
+        query_prefix = ''
+        packages = (
+            "cabextract 7-zip samba416 "  # wine
+            "wget gsed gnugrep gawk curl "  # other
+        )  # noqa: E501
+        logos_9_packages = ""  # FIXME: Missing Logos 9 Packages
+        incompatible_packages = ""  # appimagelauncher handled separately
     elif os_name == "org.freedesktop.platform":
         # Flatpak
         # Dependencies are managed by the flatpak
--- a/ou_dedetai/utils.py
+++ b/ou_dedetai/utils.py
@@ -7,7 +7,6 @@ import logging
 import os
 import queue
 import sqlite3
-import inotify.adapters # type: ignore
 import psutil
 import re
 import shutil
@@ -17,6 +16,7 @@ import subprocess
 import sys
 import tarfile
 import time
+import threading
 from ou_dedetai.app import App
 from packaging.version import Version
 from pathlib import Path
@@ -27,6 +27,10 @@ from . import network
 from . import system
 from . import wine
 
+if sys.platform.startswith("linux"):
+    import inotify.adapters  # type: ignore
+elif sys.platform.startswith(("darwin", "freebsd", "dragonfly", "netbsd", "openbsd")):
+    import select
 
 def get_calling_function_name():
     if 'inspect' in sys.modules:
@@ -659,10 +663,6 @@ def watch_db(path: str, sql_statements: list[str]):
 
     This function may run infinitely, spawn it on it's own thread
     """
-    # Silence inotify logs
-    logging.getLogger('inotify').setLevel(logging.CRITICAL)
-    i = inotify.adapters.Inotify()
-    i.add_watch(path)
 
     def execute_sql(cur):
         # logging.debug(f"Executing SQL against {path}: {sql_statements}")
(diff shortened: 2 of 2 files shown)


thw26:
After building Python 3.12, I was able to make Ou Dedetai run with TK but not with curses. Attempting to run via curses spits out this error. I am unsure why this is happening.

```
/scripts/run_app.py
  File "/home/thwright/Dev/LogosLinuxInstaller/./scripts/run_app.py", line 14, in <module>
    sys.exit(ou_dedetai.main.main())
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thwright/Dev/LogosLinuxInstaller/ou_dedetai/main.py", line 469, in main
    run()
  File "/home/thwright/Dev/LogosLinuxInstaller/ou_dedetai/main.py", line 406, in run
    config.ACTION()  # run control_panel right away
    ^^^^^^^^^^^^^^^
  File "/home/thwright/Dev/LogosLinuxInstaller/ou_dedetai/main.py", line 317, in run_control_panel
    raise e
  File "/home/thwright/Dev/LogosLinuxInstaller/ou_dedetai/main.py", line 302, in run_control_panel
    curses.wrapper(tui_app.control_panel_app)
  File "/opt/lib/python3.12/curses/__init__.py", line 94, in wrapper
    return func(stdscr, *args, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thwright/Dev/LogosLinuxInstaller/ou_dedetai/tui_app.py", line 1140, in control_panel_app
    TUI(stdscr).run()
  File "/home/thwright/Dev/LogosLinuxInstaller/ou_dedetai/tui_app.py", line 344, in run
    self.display()
  File "/home/thwright/Dev/LogosLinuxInstaller/ou_dedetai/tui_app.py", line 305, in display
    self.active_screen.display()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'display'
```

thw26:
This is able to get to the installer but it is stuck at:

"Please wait while Setup installs Logos Bible Study".

It hangs here and never proceeds. I'm not seeing any errors. It's simply as if it can't.

The GhostBSD repos have wine 10.0-devel rc4 at present. We will need to disable AppImage access, much like #321.

thw26:
A draw error occurs in the TUI in which when the next screen is meant to be drawn, it just blanks the screen.

thw26:
Further testing shows that I can more easily install now given that Python 3.12 is available, but the install process gets stuck on `wine64 wineboot --init` and it causes my VM to crash.

thw26:
https://www.phoronix.com/news/Wine-11.14-Released

This might explain the problem. We need to try again with Wine 11.14 on FreeBSD.

EDIT: Building Wine 11.14 on a GhostBSD VM.

EDIT: Testing on Wine 11.14 gets me to the "Installing Logos Bible Study" window of the Wine program, but its status never progresses. I am essentially back to https://github.com/FaithLife-Community/OuDedetai/pull/219#issuecomment-2673613025.

The install is failing at a step in the MSI launching the `CreatePendingFolder` custom action by Logos in the installer. Based on a diagnostic chat with AI, it suggested:

>Given that this is Ou Dedetai installing Logos, my leading suspicion is now:
>Wine 11.14's new FreeBSD WoW64 support can run the 64-bit MSI engine, but the installer invokes a custom action DLL that is not working correctly in this environment.

This requires evaluating and investigating this custom PE32 DLL further.

Issue #220, opened 2024-10-26
Updating ICU version fails


ctrlaltf24:
## Reproduction:
- Already have oudedetai installed with an old ICU version
- ~/Downloads/icu-win.tar.gz is still in your downloads from the original install
- Run oudedetai and install icu files

Portion of `~/.local/state/FaithLife-Community/oudedetai.log`
```
2024-10-26 01:06:40 INFO: /home/user/Downloads/icu-win.tar.gz is the wrong size.
2024-10-26 01:06:40 CRITICAL: Bad file size or checksum: /home/user/Downloads/icu-win.tar.gz
2024-10-26 01:06:40 INFO: Caught SystemExit, exiting gracefully...
2024-10-26 01:06:40 DEBUG: Closing Ou Dedetai.
2024-10-26 01:06:40 DEBUG: Closing Ou Dedetai finished.
```

## Workaround
Delete `~/Downloads/icu-win.tar.gz` and try again

PR #221, opened 2024-10-26
fix: re-download the file if the checksums don't match

ctrlaltf24:
Preserve the original behavior where it will try first to resume the download, but if it fails, it'll try again.

Needed for if the file changes - such as in the case of icu-win.tar.gz after an update

Tested:
- put the old icu-win.tar.gz in my downloads folder and modified `~/.local/state/FaithLife-Community/ICU_Version.txt` to trick it into updating. Correctly deleted the old file and downloaded the new one
- Faked a partial download with `cat ~/icu-win.tar.gz | head -c 100 > ~/icu-win.tar.gz` and modified `~/.local/state/FaithLife-Community/ICU_Version.txt` to trick it into updating. It didn't need to delete the file in order to pass

Fixes: #220

### DIFF
--- a/ou_dedetai/network.py
+++ b/ou_dedetai/network.py
@@ -158,35 +158,8 @@ def logos_reuse_download(
     targetdir,
     app=None,
 ):
-    dirs = [
-        config.INSTALLDIR,
-        os.getcwd(),
-        config.MYDOWNLOADS,
-    ]
-    found = 1
-    for i in dirs:
-        if i is not None:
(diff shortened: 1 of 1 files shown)


n8marti:
Can you also test this against a partial download of `Logos*.msi` and of `wine-devel*.appimage`, which are presumably in the Downloads folder? Maybe run the full installer with one of those chopped files for more of an integration test?

ctrlaltf24:
superseded by #224

PR #222, opened 2024-10-26
fix: always install passively

ctrlaltf24:
For some reason non-passive installs were failing

As a user I see no need to hit a single next button on a dialog I've already given my consent to install by installing our installer

Tested working with icu 72.1-custom+4

### DIFF
--- a/ou_dedetai/config.py
+++ b/ou_dedetai/config.py
@@ -98,7 +98,6 @@ PACKAGE_MANAGER_COMMAND_INSTALL: Optional[list[str]] = None
 PACKAGE_MANAGER_COMMAND_REMOVE: Optional[list[str]] = None
 PACKAGE_MANAGER_COMMAND_QUERY: Optional[list[str]] = None
 PACKAGES: Optional[str] = None
-PASSIVE: Optional[bool] = None
 pid_file = f'/tmp/{name_binary}.pid'
 PRESENT_WORKING_DIRECTORY: str = os.getcwd()
 QUERY_PREFIX: Optional[str] = None
--- a/ou_dedetai/main.py
+++ b/ou_dedetai/main.py
@@ -83,10 +83,6 @@ def get_parser():
         '-L', '--delete-log', action='store_true',
         help='delete the log file',
     )
-    cfg.add_argument(
-        '-P', '--passive', action='store_true',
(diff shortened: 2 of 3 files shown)


n8marti:
I'm happy with always installing passively. It's what I always do. I can't imagine there are many people who *want* to have to click that they agree to the EULA or who *want* to choose an alternative installation location. However, @thw26, can you see any potential issues from the FaithLife side if users don't actually have to click to accept the EULA? Or from the users' side if they some day find out they have a problem with the EULA that they never saw during install?

n8marti:
The ICU files need to be installed sometime between wine init and running the Logos installer MSI because running the MSI depends on the ICU files already being "correct". Confirmed by taking the following steps in a clean install:
```
python -m ou_dedetai.main --install-app -DF  # Logos installer fails with "windows update" error
mv ~/.local/state/FaithLife-Community/ICU_Version.txt{,.bak}  # have to avoid the ICU config check
# Go to main.py and comment out L417 that requires the app to be installed in order to install ICU files.
INSTALLDIR=~/LogosBible10 python -m ou_dedetai.main --install-icu -D  # have to pass INSTALLDIR so that ICU knows where to put the files
python -m ou_dedetai.main --install-app -DF  # Logos installer opens correctly, no "windows update" error

``

n8marti:
As mentioned in the matrix chat, I was wrong about the problem being the timing of when the ICU files are installed, but rather it seems like there's some bug in how `ICU_Version.txt` is being evaluated. Removing it still solved the problem, at any rate.

ctrlaltf24:
> I'm happy with always installing passively. It's what I always do. I can't imagine there are many people who _want_ to have to click that they agree to the EULA or who _want_ to choose an alternative installation location. However, @thw26, can you see any potential issues from the FaithLife side if users don't actually have to click to accept the EULA? Or from the users' side if they some day find out they have a problem with the EULA that they never saw during install?

Good point, if we decide to do this we should probably show/prompt a EULA agreement

n8marti:
since the non-passive installs seem to be failing related to ICU issues, which we expect to fix in #224, I think it's better to keep active Logos install the default. This preserves proper EULA acceptance, and most users won't be running it that often, and it's what they would expect if installing in Windows as well, and we still have the `--passive` flag available.

PR #224, opened 2024-10-29
feat: use ICU tarball name as version

ctrlaltf24:
This renames the installed icu tarball to include the version

Pros:
- removes need for state file
- doesn't rely on the name of the released icu tarball
- similar handling to how we download the Logos msi installer today

Cons:
- can't think of any

Fixes: #220

cc: @n8marti and @thw26

### DIFF
--- a/ou_dedetai/network.py
+++ b/ou_dedetai/network.py
@@ -14,6 +14,8 @@ from time import sleep
 from urllib.parse import urlparse
 from xml.etree import ElementTree as ET
 
+from ou_dedetai import wine
+
 from . import config
 from . import msg
 from . import utils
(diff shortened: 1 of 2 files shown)


n8marti:
What makes the ICU files a new scenario is that we can't easily compare the *installed* files with the latest release, since one is uncompressed and consists of multiple files spread around in multiple places, while the other is a single compressed file. I think the closest we could do was see if the correct .tar.gz file was downloaded, but that method wouldn't guarantee that it was the version of ICU files actually installed in the wine prefix.

ctrlaltf24:
The early return was removed, this always extracts the tarball

thw26 (approved):

PR #227, opened 2024-11-13
chore: remove unused password prompts

ctrlaltf24:
119 lines of unused code

If we want to get it back we can use git history

Didn't test, checked all functions for usages before removing. May make sense to wait for the testing framework before merging this.


ctrlaltf24:
code rotted.

Issue #230, opened 2024-11-14
installer fails to install desktop files on fedora-40


ctrlaltf24:
Fails to create desktop files on Fedora-40
```
2024-11-14 15:21:04 INFO: Creating launcher shortcuts…
2024-11-14 15:21:04 INFO: Creating desktop launcher at /home/user/.local/share/applications/LogosBible.desktop.
Exception in thread <function ensure_launcher_shortcuts at 0x7aac8cbf6ca0>:
Traceback (most recent call last):
  File "threading.py", line 1075, in _bootstrap_inner
  File "threading.py", line 1012, in run
  File "ou_dedetai/installer.py", line 681, in ensure_launcher_shortcuts
  File "ou_dedetai/installer.py", line 871, in create_launcher_shortcuts
  File "ou_dedetai/installer.py", line 781, in create_desktop_file
  File "pathlib.py", line 1013, in open
```

PR #231, opened 2024-11-14
fix: ensure local desktop file directory exists

ctrlaltf24:
Workaround:
```
mkdir -p ~/.local/share/applications
```

Fixes: #230


n8marti (changes requested):
We also need the `exist_ok=True` paramenter here. Then we can merge this straight in.

n8marti (approved):

PR #229, opened 2024-11-14
fix: add required fuse package to fedora

ctrlaltf24:
"fuse" package needed for fusermount
"fuse3" package provides fusermount3

When the fuse package isn't installed:
```
$ ./oudedetai --install-app
...
Installing registry file: /tmp/LBS.2r05magl/disable-winemenubuilder.reg
2024-11-14 14:57:56 ERROR: Error occurred in run_command() while executing "['/home/user/LogosBible10/data/bin/wine64', 'reg', 'query', 'HKCU\\Software\\Wine\\Fonts', '/v', 'Codepages']": Command '['/home/user/LogosBible10/data/bin/wine64', 'reg', 'query', 'HKCU\\Software\\Wine\\Fonts', '/v', 'Codepages']' returned non-zero exit status 127.
2024-11-14 14:57:56 WARNING: Failed to get registry value: HKCU\Software\Wine\Fonts\Codepages
2024-11-14 14:57:56 ERROR: wine.wine_proc: wine.get_registry_value returned None.
Failed to install reg file: /tmp/LBS.2r05magl/disable-winemenubuilder.reg
2024-11-14 14:57:56 CRITICAL: Failed to install reg file: /tmp/LBS.2r05magl/disable-winemenubuilder.reg
Killed
$ /home/user/LogosBible10/data/bin/wine64 reg query HKCU\\Software\\Wine\\Fonts /v Codepages
fuse: failed to exec fusermount: No such file or directory

Cannot mount AppImage, please check your FUSE setup.
You might still be able to extract the contents of this AppImage 
if you run it with the --appimage-extract option. 
See https://github.com/AppImage/AppImageKit/wiki/FUSE 
for more information
open dir error: No such file or directory
```

This using Fedora-40

PR #232, opened 2024-11-18
make dependencies lists more precise

n8marti:
- fix #229
- update dependency package names for newer debian-based distros
- remove `cabextract` and `sed` dependencies b/c they're included in the wine appimage

I will test this in debian/ubuntu/fedora VMs before removing "draft" status. I trust others will test other distros as needed :wink:.

EDIT: Explicit requirements seem to be:
- `fusermount` command
- *some* kind of `libfuse*.so`, but must be compatible with the `fusermount` binary

This is complicated because Ubuntu symlinks `/bin/fusermount` to `/bin/fusermount3`, so it only needs the `fuse3` package, which provides both `fusermount` and brings in `libfuse3-3` as its own dependency, which is the library compatible with that `fusermount` (which is really`fusermount3`). But then other distros (e.g. fedora) keep fuse 2 and 3 completely separate, so in order to provide `fusermount`, fedora users must have `fuse` installed, which in turn *needs* `libfuse.so.2` provided by `fuse-libs` but is not an explicit dependency, so both `fuse` and `fuse-libs` are explicitly given in my changes below.

### DIFF
--- a/ou_dedetai/system.py
+++ b/ou_dedetai/system.py
@@ -334,9 +334,15 @@ def get_package_manager():
         # NOTE: Package names changed together for Ubuntu 24+, Debian 13+, and
         # derivatives. This does not include an exhaustive list of distros that
         # use 'apt', so others will have to be added as users report issues.
+        # Ref:
+        # - https://askubuntu.com/a/445496
+        # - https://en.wikipedia.org/wiki/Linux_Mint
+        # - https://en.wikipedia.org/wiki/Elementary_OS
+        # - https://github.com/which-distro/os-release/tree/main
         if (
-            (config.OS_NAME == 'ubuntu' and major_ver >= '24')
-            or (config.OS_NAME == 'debian' and major_ver >= '13')
+            (config.OS_NAME == 'debian' and major_ver >= '13')
+            or (config.OS_NAME == 'ubuntu' and major_ver >= '24')
+            or (config.OS_NAME == 'linuxmint' and major_ver >= '22')
             or (config.OS_NAME == 'elementary' and major_ver >= '8')
         ):
             config.PACKAGES = (
@@ -371,7 +377,7 @@ def get_package_manager():
             "fuse2 "  # appimages
             "samba wget "  # wine
             "7zip "  # winetricks
-            "curl gawk grep patch "  # other
+            "curl gawk grep "  # other
         )
         config.L9PACKAGES = ""  # FIXME: Missing Logos 9 Packages
         config.BADPACKAGES = ""  # appimagelauncher handled separately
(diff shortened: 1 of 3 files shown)


thw26:
Won't removing cabextract and sed make using the distro's packaged wine not work?

n8marti:
if the user has installed their distro's wine, then *it* will (or should) bring in those dependencies. We shouldn't be preemptively installing them *just in case*. The default install (using the wine appimage) doesn't need them.

n8marti:
Well, it turns out ubuntu 24.04 doesn't even need `libfuse2t64`. The appimage works fine with only `libfuse3-3` being installed, and that can't be removed without also removing `ubuntu-desktop-minimal` and `grub-*`. Similar with debian. I need to test fedora 41 next. Arch (& I assume manjaro) are still needing `fuse2`.

ou_dedetai/system.py
```diff
@@ -314,15 +314,36 @@ def get_superuser_command():
...
+            "binutils wget winbind "  # wine
+            "p7zip-full "  # winetricks
+        )
+        # NOTE: Package names changed together for Ubuntu 24+, Debian 13+, and
+        # derivatives. This does not include an exhaustive list of distros that
+        # use 'apt', so others will have to be added as users report issues.
+        if (
+            (config.OS_NAME == 'ubuntu' and major_ver >= '24')
+            or (config.OS_NAME == 'debian' and major_ver >= '13')
+            or (config.OS_NAME == 'elementary' and major_ver >= '8')
```
thw26, line 340: Linux Mint would be another we should add due to its popularity with new users

ou_dedetai/system.py
```diff
@@ -355,7 +381,12 @@ def get_package_manager():
         config.PACKAGE_MANAGER_COMMAND_REMOVE = ["pamac", "remove", "--no-confirm"]  # noqa: E501
         config.PACKAGE_MANAGER_COMMAND_QUERY = ["pamac", "list", "-i"]
         config.QUERY_PREFIX = ''
-        config.PACKAGES = "patch wget sed grep gawk cabextract p7zip samba bc libxml2 curl"  # noqa: E501
+        config.PACKAGES = (
+            "fuse2 "  # appimages
+            "samba wget "  # wine
+            "p7zip "  # winetricks
+            "bc curl gawk grep libxml2 patch "  # other
```
thw26, line 388: Remove bc, libxml2, patch

ou_dedetai/system.py
n8marti: done

ou_dedetai/system.py
n8marti: and done

thw26:
This looks good. Only question is the large chunk of text that says to be removed once verified. Is that a final change before merge?

n8marti:
> This looks good. Only question is the large chunk of text that says to be removed once verified. Is that a final change before merge?

I've now tested the binary in clean fedora 41 and clean ubuntu 24.04 and everything works well. I'm going to remove the comments and merge to `main`.

PR #242, opened 2024-12-10
feat: pass LANG through to wine

ctrlaltf24:
Allows both Logos' MSI installer and the product to default to other languages if supplied

### DIFF
--- a/ou_dedetai/wine.py
+++ b/ou_dedetai/wine.py
@@ -584,6 +584,10 @@ def get_wine_env():
     # if config.LOG_LEVEL > logging.INFO:
     #     wine_env['WINETRICKS_SUPER_QUIET'] = "1"
 
+    locale = os.getenv("LANG")
+    if locale:
+        wine_env["LANG"] =  locale
+
     # Config file takes precedence over the above variables.
     cfg = config.get_config_file_dict(config.CONFIG_FILE)
     if cfg is not None:


n8marti (commented):
Does this have any effect on the Logos app? It seems like it could make `wine.log` output potentially difficult to follow for us devs. That being said, my `wine.log` already includes lines of output in French since my system is in French, so I'm wondering what practical effect this would actually have. Have you tested it?

ctrlaltf24:
Yes it has the effect stated in the comment - setting the logos language 

I haven't checked the language of the logs but generally application logs aren't internationalized (as they're mostly for the original developers)

n8marti:
Looking more closely at the code, `LANG` *is* already passed through to `wine`. Just above in the same function (L564) the os's environment is *copied*, then wine-specific variables are set on top of that. Indeed, when I run an install, the wine init window is already in French. However the MSI installer is in English, and then winetricks messages are in French, presumably because it's just bash commands and their output has all been localized.

n8marti:
If I try an install forcing the MSI language with ProductLanguage its interface still uses English. Maybe it only includes English, or maybe French is not included but others might be? I modified the msiexec line in `scripts/quick-install.sh` to be this:
```bash
$WINE msiexec /i "$DOWNLOADS/$LOGOS_MSI_NAME" ProductLanguage=1036 || exit 1
```
Then ran the script:
```
$ export LOGOS_RELEASE_VERSION=38.1.0.0002
$ export LOGOS_WINEPREFIX=~/wine-logos
$ export LOGOS_WINE=/opt/wine-devel/bin/wine64
$ ./g/LogosLinuxInstaller/scripts/quick-install.sh -y
```

Ref:
- https://stackoverflow.com/a/7725197
- https://learn.microsoft.com/en-us/windows/win32/msi/localizing-the-error-and-actiontext-tables

n8marti:
I'm going to close this as per https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/242#issuecomment-2604813166

Feel free to reopen if necessary.

PR #257, opened 2025-01-17
fix: remove pacman --overwrite flag

### LINKED ISSUE #235: Remove --overwrite flag from pacman command to install dependencies
When running `oudedetai` for the first time on Arch, it asks you to install the necessary dependencies with
```pacman -Syu --overwrite \* --nocomfirm --needed```

From the [Arch Wiki](https://wiki.archlinux.org/title/System_maintenance#Avoid_certain_pacman_commands):

> Generally avoid using the --overwrite option with pacman. The --overwrite option takes an argument containing a glob. When used, pacman will bypass file conflict checks for files that match the glob. In a properly maintained system, it should only be used when explicitly recommended by the Arch Developers. See the [#Read before upgrading the system](https://wiki.archlinux.org/title/System_maintenance#Read_before_upgrading_the_system) section.

Is there any good reason for prompting the user to use the `--overwrite` flag?  If the user has a properly maintained system, there should be no need for the `--overwrite` flag.  And if their system is not properly maintained, then they probably have bigger issues they'll need to deal with.

### PR DESCRIPTION
**Pull request by thw26**
Fixes #235

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,16 @@
 # Changelog
 
 - 4.0.0-beta.6
+  - Fix #234 [N. Shaaban]
+  - Fix #187 [N. Shaaban]
+  - Fix #168 [N. Shaaban]
+  - Fix #155 [N. Shaaban]
+  - Fix #147 [N. Shaaban]
+  - Fix #253 [N. Marti]
+  - Fix #252 [N. Marti]
 	- Implemented network cache [N. Shaaban]
 	- Refactored configuration handling (no user action required) [N. Shaaban]
+  - Fix #235 [T. H. Wright]
 - 4.0.0-beta.5
 	- Filter out broken Logos v39 release [N. Shaaban]
 	- Fix #17 [T. H. Wright]
--- a/ou_dedetai/system.py
+++ b/ou_dedetai/system.py
@@ -474,7 +474,7 @@ def get_package_manager() -> PackageManager | None:
         )
         incompatible_packages = ""  # appimagelauncher handled separately
     elif shutil.which('pacman') is not None:  # arch, steamOS
-        install_command = ["pacman", "-Syu", "--overwrite", "\\*", "--noconfirm", "--needed"]  # noqa: E501
+        install_command = ["pacman", "-Syu", "--noconfirm", "--needed"]  # noqa: E501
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: ou_dedetai/system.py [hunk 1]
```diff
@@ -917,4 +917,4 @@ def check_incompatibilities(app: App):
             f"For more information, see: {constants.REPOSITORY_LINK}/issues/114"
         )
         app.approve_or_exit(question_text, secondary)
-        remove_appimagelauncher(app)
\ No newline at end of file
+        remove_appimagelauncher(app)
```
thw26, line 920: Not sure what this change is doing here. Reviewing…

thw26: This code exists in the main file. Not sure what this change is getting at. I added a PEP8 extra line while reviewing, but unsure what else to make of this change.

n8marti: It looks like it's a warning that there's no newline at the end of the file, based on what [this](https://www.reddit.com/r/learnprogramming/comments/t7qofi/why_does_github_care_if_a_file_doesnt_have_a/) says. But I don't know why it would highlight as a change?

n8marti (approved):
I saw the request to remove the `overwrite` option, but I'm not familiar with the wisdom of the decision. You obviously correctly make it happen with this minor change, though! Looks good to me.

CHANGELOG.md
```diff
@@ -1,8 +1,16 @@
...
+  - Fix #234 [N. Shaaban]
+  - Fix #187 [N. Shaaban]
+  - Fix #168 [N. Shaaban]
+  - Fix #155 [N. Shaaban]
+  - Fix #147 [N. Shaaban]
+  - Fix #253 [N. Marti]
+  - Fix #252 [N. Marti]
 	- Implemented network cache [N. Shaaban]
 	- Refactored configuration handling (no user action required) [N. Shaaban]
+  - Fix #235 [T. H. Wright]
```
ctrlaltf24, line 13: nit: difference of tabs and spaces

Who's the intended audience for the changelog? Generally it's nice in a changelog to be human readable (linking to issues for additional context)

thw26: This format is a holdover from the bash script. We can look to renovate the format perhaps in a new PR.

PR #260, opened 2025-01-17
add quick-install|run scripts for easier testing

n8marti:
The purpose of these two scripts is to facilitate easy testing of the wine setup.
- `scripts/quick-install.sh` installs and configures the wine system
- `scripts/quick-run.sh` launches the installed Logos app

There is currently very little data validation, and the user is responsible for setting up the host system and downloading all necessary files to `~/Downloads` beforehand. The scripts rely on 3 environment variables, which are requested interactively if not already set:
- `LOGOS_RELEASE_VERSION`: e.g. 38.1.0.0002
- `LOGOS_WINEPREFIX`: e.g. `~/LogosBible10`
- `LOGOS_WINE`: e.g. `/opt/wine-devel/bin/wine64`

### DIFF
--- a/scripts/quick-install.sh
+++ b/scripts/quick-install.sh
@@ -1,0 +1,98 @@
+#!/usr/bin/env bash
+
+# This script assumes you've managed all the system prerequisites yourself.
+# It can be run non-interactively if you prepare a few things beforehand:
+# 1. Set env varaibles. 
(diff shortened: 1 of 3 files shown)


thw26 (approved):
My only comments are usability issues covered in the commented section. Otherwise it works.

PR #261, opened 2025-01-17
feat: disable automatic updates

ctrlaltf24:
Edits Logos' internal dbs both once on startup and every time the db is written to:
- removing the entries saying there is an application update
- enforcing "Automatically start downloads" to false

Also recovers failed upgrades

Further consideration is needed whether or not enforcing "Automatically start downloads" is necessary for a stable system


ou_dedetai/utils.py
```diff
@@ -683,8 +682,69 @@ def stopwatch(start_time=None, interval=10.0):
...
+            if swallow_one:
+                swallow_one = False
+                continue
+            execute_sql(cur)
+            swallow_one = True
+    # Shouldn't be possible to get here, but on the off-chance it happens, 
+    # we'd like to know and cleanup
+    logging.debug(f"Stopped watching {path}")
+    cur.close()
+    con.close()
```
n8marti, line 750: How about using context managers here? Something like what I did [here](https://github.com/n8marti/logos-resource-updater/blob/e39ebcac86c63462a785d9a2ffdcba4058774919/src/logos_resource_updater/logosmgr.py#L19).

ou_dedetai/cli.py
```diff
@@ -68,6 +70,10 @@ def run_indexing(self):
 
     def run_installed_app(self):
         self.logos.start()
+        # Keep the process running so that our background threads can keep running
+        while self.logos.logos_state != LogosRunningState.STOPPED:
+            time.sleep(10)
+            self.logos.monitor()
```
n8marti, line 76: Can this be achieved instead by using `daemon=False` on the long-running threads?

ou_dedetai/config.py
```diff
@@ -789,6 +789,14 @@ def installer_binary_dir(self) -> str:
...
         return f"{self.install_dir}/data/bin"
 
+    @property
+    def logos_appdata_dir(self) -> Optional[str]:
+        """Path to the user's Logos installation under AppData"""
+        wine_user = self.wine_user
+        if not wine_user:
+            return None
+        return self.wine_prefix + f'/drive_c/users/{wine_user}/AppData/Local/Logos'
+
```
n8marti, line 799: I think this property can be used in the backup/restore code, as well. Can you take a look (`control.py`)?

ou_dedetai/cli.py
ctrlaltf24: Don't think so, if the CLI process exits the the threads spawned by python terminate

ou_dedetai/config.py
ctrlaltf24: Yup line 71, good catch

n8marti (approved):
The code looks good, and I've been assured specifically that the recovery feature has been tested in both TUI and GUI.

PR #263, opened 2025-01-20
fix: Support for Logos 39+ via Mst that changes msi config

jg00dman:
This adds an mst which modifies 4 attributes of the msi installer. The mst is applied at the time the msi is installed if the Logos version is 39 or above. The mst should remain compatible with future releases unless the msi tables change. Then we'll need a new mst.

Screenshot_20250116_201616

Basically we are allowing the LogosStub install to fail because it is in msix format and isn't needed at runtime on linux. LogosStub just forwards ref.ly urls to logos so if you click on those links logos will open on windows.


ou_dedetai/network.py
```diff
@@ -569,7 +569,7 @@ def _get_faithlife_product_releases(
         #    break
 
     # Logos 39+ doesn't install
-    filtered_releases = utils.filter_versions(releases, 39, 1)
+    filtered_releases = utils.filter_versions(releases, 40, 1) #jg
     logging.debug(f"Available releases: {', '.join(releases)}")
     logging.debug(f"Filtered releases: {', '.join(filtered_releases)}")
```
ctrlaltf24, line 575: If you wanted to stop it filtering, comment out these lines, then change the return below to 'releases' rather than filtered_releases

ou_dedetai/wine.py
```diff
@@ -271,11 +272,32 @@ def disable_winemenubuilder(app: App, wine64_binary: str):
...
+        # Copy the MST file if it doesn't already exist
+        if not mst_destination.is_file():
+            shutil.copy(mst_source, mst_destination)
+        logging.debug(f"MST present: {mst_destination.is_file()}")
+
+        # Hard-code the Windows path for the MST file
+        transform_winpath = r"C:\LogosStubFailOK.mst"
+        exe_args.append(f'TRANSFORMS="{transform_winpath}"')
+        logging.debug(f"Hard-coded TRANSFORMS path added: {transform_winpath}")
+
```
ctrlaltf24, line 299: nit: should we delete the mst from the wine bottle here? Little hit ugly to leave things in top level C

scripts/ensure-venv.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 python_ver=$(wget -qO- https://www.python.org/ftp/python/ | grep -oE '3\.12\.[0-9]+' | sort -u | tail -n1)
-prefix=/opt
+prefix=
```
ctrlaltf24, line 4: Looks like this is an accidental change, can you put it back to opt please?

ctrlaltf24 (commented):
Looks good! Well done John! Couple nits (aka stuff I'd like to see but it can be merged into main without them)

ou_dedetai/wine.py
jg00dman: could do but we'd need to recreate it each time we run an update.

scripts/ensure-venv.sh
jg00dman: yes had to do it to make it work on kubuntu

ou_dedetai/wine.py
jg00dman: done

ou_dedetai/network.py
jg00dman: done

scripts/ensure-venv.sh
jg00dman: done

ctrlaltf24 (approved):

ou_dedetai/wine.py
jg00dman: this approach didn't work... seems there is some threading which means the script deleted it before it could be used. I've moved it to the clean up step at the end of the process and it now works nicely.

ou_dedetai/installer.py
```diff
@@ -289,6 +290,14 @@ def ensure_product_installed(app: App):
             process.wait()
 
     # Clean up temp files, etc.
+    mst_destination = Path(app.conf.install_dir) / "data/wine64_bottle/drive_c/LogosStubFailOK.mst"
```
ctrlaltf24, line 293: Why was this moved here?

The place in the code that a temporary file is created should be the place where it's cleaned up

ctrlaltf24:
Tested clean install (up to showing the splash screen, before login), works version v39 shown

Would still like to see https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/263#discussion_r1923220250 and https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/263#discussion_r1923344024 resolved, however it can be merged in this state

ou_dedetai/constants.py
```diff
@@ -1,10 +1,23 @@
...
 # This is relative to this file itself
 APP_IMAGE_DIR = Path(__file__).parent / "img"
 
+# Are we running from binary or src?
+if getattr(sys, 'frozen', False):
+    # We are running inside a PyInstaller bundle
+    BUNDLE_DIR = Path(sys._MEIPASS)
+else:
+    # We are running in normal development mode
+    BUNDLE_DIR = Path(__file__).resolve().parent
```
n8marti, line 16: This same check already happens in `system.py` [L516](https://github.com/jg00dman/LogosLinuxInstaller/blob/f40323c69159c22e9815fb266acd6581a221c495/ou_dedetai/system.py#L516); e.g.
```python
from system import get_runmode
if get_runmode() == 'binary':
    BUNDLE_DIR = Path(sys._MEIPASS)
else:
    BUNDLIE_DIR = Path(__file__).resolve().parent
```

ou_dedetai/wine.py
```diff
@@ -271,13 +272,35 @@ def disable_winemenubuilder(app: App, wine64_binary: str):
...
     wine_exe = app.conf.wine64_binary
-    exe_args = ["/i", f"{app.conf.install_dir}/data/{app.conf.faithlife_installer_name}"] #noqa: E501
+    exe_args = ["/i", f"{app.conf.install_dir}/data/{app.conf.faithlife_installer_name}"]  # noqa: E501
+
+    # Add passive mode if specified
     if app.conf._overrides.faithlife_install_passive is True:
-        exe_args.append('/passive')
+        exe_args.append("/passive")
+
+    # Add MST transform if needed
```
n8marti, line 283: Does this MST transform work in such a way that the _only_ change to the way the MSI works is that the MSI is able to continue if the specific MSIX file in question fails? I'm wondering if it's worth _always_ running the MST. That would be nice if there's no negative effect to any earlier installers, nor would it have any negative effects to any future installers that _don't_ use MSIX (or if wine starts providing compatibility).

ctrlaltf24: It seems logical to keep the v39 check while that question is up in the air

jg00dman: I don't know someone would need to try an earlier msi with the mst. don't have time sorry.

ou_dedetai/constants.py
jg00dman: yeah but it is way later than I need the info. I guess it could get reused from here later?

ou_dedetai.spec
```diff
@@ -6,7 +6,7 @@ a = Analysis(
     pathex=[],
     #binaries=[('/usr/bin/tclsh8.6', '.')],
     binaries=[],
-    datas=[('ou_dedetai/img/*icon.png', 'ou_dedetai/img')],
+    datas=[('ou_dedetai/img/*icon.png', 'ou_dedetai/img'),('assets/LogosStubFailOK.mst', 'assets')],
```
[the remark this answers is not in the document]
jg00dman: trust your judgement on that one.

ou_dedetai/constants.py
ctrlaltf24: Import from system here would create a circular import. This is fine.

ctrlaltf24:
Tested clean installation, MSI installs successful with v39 https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/268

Also tested recovering a broken v38->v39 install, successfully recovered v39

PR #267, opened 2025-01-22
Remove winetricks from default install

### LINKED ISSUE #170: Simplify winetricks choices

### LINKED ISSUE #177: Winetricks Error
While testing #164, I discovered this error in the new wine log:

```
Executing cabextract -q -d /home/thwright/LogosBible10/data/wine64_bottle/dosdevices/c:/windows/temp /home/thwright/.cache/winetricks/tahoma/IELPKTH.CAB
/home/thwright/.cache/winetricks/tahoma/IELPKTH.CAB: WARNING; possible 5592 extra bytes at end of file.
rm: remove write-protected regular file '/home/thwright/LogosBible10/data/wine64_bottle/dosdevices/c:/windows/Fonts/tahoma.ttf'?
```

This must be causing some kind of hang. Since my work on this branch, I have noticed several issues that cause the program to hang. I have fixed at least one of two, and this might be a third.

### LINKED ISSUE #254: Why not automatically default to latest winetricks version?
165dad8 switched to using `winecfg` instead of `winetricks` to set the windows version. It's unclear exactly why, but this resolved several issues:
- there were sometimes zombie wine processes after closing the app (partial resolution)
- Logos wouldn't start after install
- setting the windows version goes 10x faster

But not knowing why the winetricks command was not working properly makes me wonder if our 2022 winetricks version is just too old. Setting a minimum version is good as long as we're willing to maintain it, but why not just default to the latest release instead? The script is not that big, so we could even consider packaging it into oudedetai ourselves.

To implement this, rather than maintain a WINETRICKS_VERSION constant, we would check the winetricks GitHub releases list and get the latest version string from there.

### LINKED ISSUE #258: Cache Fonts
In addition to having the `-F` option, could we cache fonts downloaded through Winetricks to speed up subsequent installs?

FWIW, I install my fonts to my .fonts dir (symlinked to a Nextcloud syncing dir).

n8marti:
Fixes #254, making it redundant. Closes #170, closes #177, closes #258.
Needs testing for regressions; specifically:
- [x] font display issues due to lack of `corefonts`?
- [x] Are winemenubuilder entries created in non-appimage wine?
- [x] ~font display issues due to lack of `fontsmoothing="rgb"`?~ (edit: now included)
- [x] ~display issues due to lack of `renderer="gdi"`?~ (edit: now included)
- [x] Any sign of D3D compiler crashes?


thw26:
We also need to test if the d3dcompiler crashes I saw have returned or not.

- [ ] d3dcompiler crashes return?

n8marti:
Yeah, we need to be sure that D3D isn't a problem, either. How do we test that? I kind of assumed just starting the app would cause it to crash if D3D wasn't handled correctly.

Note: This PR will make #254 redundant.

n8marti:
Not knowing what all to test, this is what I tried so far:

Feature | compatibility
:-- | :-:
Atlas: zoom & pan | +
Bible: open w/ interlinear | +
Bible: synchronized scrolling | +
Bible: word study, Greek: open | +
Layout: open | +
Library: open | +
Menu: open | +/- (works for some in wine 10+)
Print/Export: export to RTF | +
Search: "Person:Jesus" | +

ou_dedetai/main.py
```diff
@@ -202,9 +178,23 @@ def get_parser():
         # help='check resources'
         help=argparse.SUPPRESS,
     )
+    cmd.add_argument(
```
ctrlaltf24, line 181: Do we need these arguments (the ones to basically run wine or winetricks)? There are already syslinks in standard locations if we need to debug using them.

ou_dedetai/tui_app.py
```diff
@@ -476,17 +476,16 @@ def choice_processor(self, stdscr, screen_id, choice):
...
-            20: self.win_ver_logos_select,
-            21: self.win_ver_index_select,
-            24: self.confirm_restore_dir,
-            25: self.choose_restore_dir,
+            11: self.logos.start,
+            12: self.waiting_finish,
+            13: self.waiting_resize,
+            14: self.password_prompt,
+            15: self.utilities_menu_select,
+            16: self.renderer_select,
```
ctrlaltf24, line 484: These numbers are used elsewhere in the code as indexes, ideally these should be moved to constants, however in the meantime making sure the numbers don't change (even if there is gaps) is good enough for now.

ou_dedetai/wine.py
```diff
@@ -348,39 +360,25 @@ def run_wine_proc(
 
     cmd = f"subprocess cmd: '{' '.join(command)}'"
     logging.debug(cmd)
+    if winecmd.endswith('winetricks'):
+        encoding = 'cp437'
+    else:
+        encoding = 'utf-8'
```
ctrlaltf24, line 366: Is there any case where this isn't true? The logic was more complex before, was that for a reason?

ou_dedetai/wine.py
```diff
@@ -392,28 +390,11 @@ def run_winetricks(app: App, *args):
...
     else:
         logging.debug('<None>')
-
-
-def install_d3d_compiler(app: App):
-    cmd = ['d3dcompiler_47']
-    run_winetricks(app, *cmd)
-
-
-def install_fonts(app: App):
```
ctrlaltf24, line 408: Isn't installing fonts needed? Maybe it's not needed in the appimage, but isn't it needed if a different wine binary is used?

ctrlaltf24 (dismissed):

ou_dedetai/wine.py
n8marti: Tahoma is the one we were installing explicitly, and it's included in both the appimages we're using as well as winehq builds:
```
$ dpkg -L wine-devel | grep tah
/opt/wine-devel/share/wine/fonts/tahoma.ttf
/opt/wine-devel/share/wine/fonts/tahomabd.ttf
```
@thw26 has mentioned that `corefonts` were maybe needed in the past, but I've looked at Greek and Hebrew texts in Logos, and they without `corefonts` being installed. Maybe we need to try other less common scripts, such as Arabic or Chinese? I'm open to suggestions, but I want to see evidence that they're needed before deciding we need to keep installing them.

ou_dedetai/wine.py
n8marti: Yes, the more complicated logic was needed to circumvent some terminal output errors in the past. However, once we switched to `run_wine_proc` sending stdout and stderr directly to wine.log, all that code starting with `with process.stdout` (old L363) was completely bypassed anyway. And no one has complained of encoding errors in the wine.log file. Either it's irrelevant now, or too obscure to have been noticeably recreated.

ou_dedetai/tui_app.py
n8marti: Oh, I'll put those back, then. I didn't realize the indexes weren't simply sequential.

ou_dedetai/main.py
n8marti: Having them as command args makes it much easier to run them within the correct wine environment. If someone wants to run them directly from the terminal they have to first set `WINEPREFIX`, possibly `WINELOADER`, `WINE`, and `WINESERVER`, and potentially `WINEDLLOVERRIDES` in the env, depending on what they are trying to accomplish.

ou_dedetai/wine.py
n8marti: I'm also able to open and see the text of the "ERV: Arabic Bible" without any character issues that I can see.

thw26:
I'm working on a few other things right now, but I want to see this merged soon. Give me some time and I will run a second install so that I can check the Aramaic crashes I would regularly see before d3d's install.

ou_dedetai/wine.py
thw26: I usually store the Windows fonts in my .fonts dir(symlinked elsewhere). I think I have noticed at least one crash needing corefonts. If we can provide the fonts another way though, then that's better. We could even copy them from the AppImage you mention to the user's font dir if we have access.

n8marti:
I think I should test this in a clean VM, too, to make sure I'm not "helping" things out with my own system setup. I know wine is good at finding existing system fonts and making them available to wine apps, to that's something we need to control for in testing.

ou_dedetai/main.py
ctrlaltf24: I'd almost rather support a command that exports those than wrapping wine/winetricks. For example (something like): 
```bash
source `our_binary --envs`
```

This is fine for now, we may want to consider if we want to maintain this for v1

ou_dedetai/cli.py
```diff
@@ -78,6 +69,13 @@ def run_installed_app(self):
...
 
+    def wine(self):
+        wine.run_wine_proc(
+            self.conf.wine64_binary,
+            self,
+            exe_args=(self.conf._overrides.wine_args or [])
+        )
+
     def run_winetricks(self):
         wine.run_winetricks(self)
```
ctrlaltf24, line 80: How does the CLI ensure winetricks is installed in this mode? Personally I'd be fine with (python psudo-code):

```py
from typing import Never

def run_winetricks() -> Never
  if appimage:
    winetricks_binary = [appimage, "winetricks"]
  else if found_in_path("winetricks"):
     winetricks_binary = ["winetricks"]
  else:
    app.exit("winetricks not installed")
  cmd_run(winetricks) # ideally capture stdin/stdout
  # If the process passes
  os.exit(0)
```

ctrlaltf24 (commented):
Code review looks alright, still need to test, especially TUI (making sure there is no lingering winetricks references)

n8marti:
> I think I should test this in a clean VM, too, to make sure I'm not "helping" things out with my own system setup. I know wine is good at finding existing system fonts and making them available to wine apps, to that's something we need to control for in testing.

Did some testing in clean ubuntu 24.04. Uninstalled all system font packages except 3, used winehq-stable 10.0.0 binary, and there are no signs of font issues, neither in English, nor Greek, nor Hebrew. Furthermore, I found that Logos installs `System/Logos.Fonts.dll` (~80MB) that seems to include all needed fonts anyway.

I feel convinced that winetricks fonts are not needed. Anything else we can do to test this? We could ask for community testers to try the binary in the test-builds repo on their various systems.

ou_dedetai/cli.py
n8marti: I'd be happy to drop the `--winetricks` subcommand, but I find the `--wine` one very useful, and I don't think it adds too much additional maintenance overhead.

n8marti: I'd be happy to drop the `--winetricks` subcommand, but I find the `--wine` one very useful, and I don't think it adds too much additional maintenance overhead.

ou_dedetai/installer.py
```diff
@@ -211,62 +194,24 @@ def ensure_wineprefix_init(app: App):
...
-            app.status("Disabling winemenubuilder…")
-            wine.disable_winemenubuilder(app, app.conf.wine64_binary)
-
-        if not utils.grep(r'"renderer"="gdi"', usr_reg):
-            app.status("Setting Renderer to GDI…")
-            wine.set_renderer(app, "gdi")
-
-        if not utils.grep(r'"FontSmoothingType"=dword:00000002', usr_reg):
-            app.status("Setting Font Smoothing to RGB…")
-            wine.install_font_smoothing(app)
```
ctrlaltf24, line 244: Is this done without winetricks? Don't recall a line doing this.

ctrlaltf24:
Looks like tahoma is included in the wine build https://gitlab.winehq.org/wine/wine/-/wikis/Create-Fonts#truetype-fonts

> I feel convinced that winetricks fonts are not needed. Anything else we can do to test this? We could ask for community testers to try the binary in the test-builds repo on their various systems.

Recovering is easy enough should we discover a user needs corefonts. They can simply execute: `~/LogosBible10/data/bin/selected_wine.AppImage winetricks corefonts`. I'm not sure we need to play it that safe if we already tested and everything looks good. 

Removing d3d and removal of fontsmoothing are more worrysome to me than the font changes. There is a mantra that some programmers live by: trust your past self (until proven otherwise that is). It's doubtful these were done for no reason, would be ideal to figure out why they were added in the first place before we can have confidence removing them is fine.

ou_dedetai/installer.py
n8marti: Before this PR we were doing it with winetricks, yes. The [now unused] function still exists in `wine.py` L398-401. But if we decide we want to keep this modification we can do it directly with a registry change rather than using winetricks. This is the same kind of change I made to `set_renderer_to_gdi` in `wine.py` L280-287.

n8marti:
> They can simply execute: `~/LogosBible10/data/bin/selected_wine.AppImage winetricks corefonts`

Yes, as long as they also set WINEPREFIX.

n8marti:
> Removing d3d and removal of fontsmoothing are more worrysome to me than the font changes.

Logos already includes `System/D3DCompiler_47_cor3.dll`, which it very well may not have included in v8.### mentioned in the PR you linked to. The DLL found in the runtime dir normally takes precedence over one in C:\windows\syswow64. I don't know how the `_cor3` suffix affects that, but I doubt Logos is using the C:\windows DLL anyway. (This can be tested in working installation by renaming/removing the two DLLs in system32 and syswow64 and seeing if there are any errors.) Moreover, I've just tested 2 new wineprefixes using both the 10-rc5 appimage and winehq's 10-rc5 ubuntu package, and it turns out that a plain old init'ed wineprefix *already includes d3dcompiler_47.dll* in the two C:\windows\sys* locations.

For fontsmoothing, as I mentioned [above](https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/267#discussion_r1937555471), we can still work it back in without using winetricks. I don't see any harm in that, and it's fast to run using regedit, and it might still make fonts appear cleaner than without it.

ou_dedetai/installer.py
ctrlaltf24: I'd like to still do it, I suspect we initially did it for a reason, would rather have a couple lines of python run than the installer get less stable

ctrlaltf24 (commented):
We can remove winetricks_skip and install_fonts skip from EphermalConfig class in config.py at about line 199 (and anyplace that references it)

Unresolved comments:
- https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/267#discussion_r1934854797

ou_dedetai/config.py
```diff
@@ -196,7 +194,6 @@ class EphemeralConfiguration:
     
     Some commands always send -q"""
 
-    winetricks_skip: Optional[bool]
     install_dependencies_skip: Optional[bool]
     """Whether to skip installing system package dependencies"""
     install_fonts_skip: Optional[bool]
```
ctrlaltf24, line 199: This one too

ctrlaltf24 (commented):
Still two more cleanups, comment and this:
https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/267#discussion_r1934854797

ou_dedetai/cli.py
n8marti: I've opted to remove remaining winetricks code. I've run a couple of test installs and TUI commands, but I'd like to test a bit more before asking for another review.

ctrlaltf24 (approved):

PR #271, opened 2025-01-23
fix: appimage download issues

ctrlaltf24:
Fixes issue where new appimage would be downloaded with the wrong name and moves the recommended appimage to the top of the list

Reproduction:
- install when there is an older recommended appimage
- delete LogosBible10 folder
- attempt install again

Expected behavior: new appimage downloads with correct name

Actual behavior: appimage downloads with the old name

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,18 @@
 # Changelog
 
 - 4.0.0-beta.6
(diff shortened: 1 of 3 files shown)


n8marti (approved):
This looks good. Thanks for fixing the appimage list ordering issue.

PR #282, opened 2025-01-29
Fix AppImage update procedure in GUI

### LINKED ISSUE #273: Application doesn't think 10.0 appimage is newer than 9.19
Reproduction:
- Have an installation with a v9.X wine version
- Open the latest GUI installer (now that 10.X is the latest appimage)
- Look at Update to Latest Appimage button

Expected behavior: button clickable
Actual behavior: Button grayed out

Clipping of debug logs:
```
2025-01-23 03:59:52 DEBUG: Current wine release: 9.19
2025-01-23 03:59:52 DEBUG: Recommended wine release: 10.0
```

Looks like the version comparison logic is string based ([here](https://github.com/FaithLife-Community/LogosLinuxInstaller/blob/7f5ac1113edcdb6d873748ee627aa69d75310ad4/ou_dedetai/utils.py#L394-L407)) rather than integer based.

n8marti:
Fix #273

Use `packaging.version.Version` to compare version numbers, which allows for `major.minor.hotfix` format, as well as `major.minor` format, and is in other ways more flexible than integer or float comparisons.

### DIFF
--- a/ou_dedetai/gui_app.py
+++ b/ou_dedetai/gui_app.py
@@ -478,7 +478,7 @@ class ControlWindow(GuiApp):
             command=self.update_to_latest_lli_release
         )
         self.gui.latest_appimage_button.config(
-            command=self.update_to_latest_appimage
+            command=self.start_appimage_update
         )
         self.gui.set_appimage_button.config(command=self.set_appimage)
         self.gui.get_winetricks_button.config(command=self.get_winetricks)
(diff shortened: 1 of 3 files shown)


n8marti:
This still needs to be tested, hence the "draft" status.

ou_dedetai/gui_app.py
```diff
@@ -579,9 +579,17 @@ def set_appimage_symlink(self):
...
         self.status("Updating to latest AppImage…")
-        self.start_thread(self.set_appimage_symlink)
+        self.gui.latest_appimage_button.state(['disabled'])
+        evt = '<<AppImageUpdateDone>>'
+        self.root.bind(evt, self.update_latest_appimage_button)
+        self.start_thread(self.update_to_latest_appimage, evt=evt)
+
+    def update_to_latest_appimage(self, evt=None):
+        utils.update_to_latest_recommended_appimage(self)
+        self.root.event_generate(evt)
```
ctrlaltf24, line 591: correct me if I'm wrong but don't we only need one function for updating the appimage? Do we still need set_appimage_symlink?

For a function spawning a new thread a pattern I've been using is declaring the function you want to spawn inside the function itself. For cases when you won't want to run the function outside of a separate thread. In python that has the added benefit of capturing the function's arguments. Seen below:

```py
def my_function_that_spawns_a_thread(self, arg_1):
  def _run():
    # do work
    self.function(arg_1)
    pass
  self.start_thread(_run)
```

ctrlaltf24 (approved):
Reviewed code, looks good. Lmk if you want me to test as well

PR #288, opened 2025-02-01
fix: convert MST file to windows path before use (#286)

### LINKED ISSUE #286: Beta7: Regression: Wine binary installs are not working

n8marti:
This fixes #286. It seems the appimage would work fine with the hard-coded windows path `C:\LogosStubFailOK.mst`, but this was failing when using system wine. But using `winepath -w` to convert the original MST file path to a winepath, and ensuring that the path is not quoted, succeeds. In fact, it's probably the lack of quotes that solves the problem, but this also adds a simplification in not needing to copy the MST file into the WINEPREFIX before using it.

### DIFF
--- a/ou_dedetai/installer.py
+++ b/ou_dedetai/installer.py
@@ -293,14 +293,6 @@ def ensure_product_installed(app: App):
     app.conf._installed_faithlife_product_release = None
 
     # Clean up temp files, etc.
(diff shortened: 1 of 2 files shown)


ctrlaltf24 (approved):
Code looks good, so long as it works feel free to merge!

n8marti:
> Code looks good, so long as it works feel free to merge!

Yes, tested with current recommended appimage and with system wine.

Issue #292, opened 2025-02-05
undefined symbol: rl_trim_arg_from_keyseq


ctrlaltf24:
Reproduction:
- on some Linux systems, one observed, one observed nixOS
- preform an install using the GUI
- failure with the below message

```
Ready to install!
Installing…
Asking questions if needed…
Install is running…
Ensuring installation directories…
Ensuring system dependencies are met…
Checking Logos 10 dependencies…
2025-01-30 13:03:19 ERROR: Your distro requires manual dependency installation.
Installed dependencies.
Ensuring wine AppImage is downloaded…
Verifying /home/elliots/Downloads/wine-staging_10.0-rc5-x86_64.AppImage…
Ensuring wine executables are available…
Creating wine appimage symlinks…
Copying: /home/elliots/Downloads/wine-staging_10.0-rc5-x86_64.AppImage into: /home/elliots/LogosBible10/data/bin
Ensuring winetricks executable is available…
Downloading winetricks from the Internet…
Installing winetricks v20220411…
Verifying /home/elliots/Downloads/20220411.zip…
/bin/sh: symbol lookup error: /bin/sh: undefined symbol: rl_trim_arg_from_keyseq
Exception in thread Ou Dedetai <function InstallerWindow.start_install_thread.<locals>._install at 0x7afae68540e0>:
Traceback (most recent call last):
  File "threading.py", line 1075, in _bootstrap_inner
  File "threading.py", line 1012, in run
  File "ou_dedetai/gui_app.py", line 421, in _install
  File "ou_dedetai/installer.py", line 358, in install
  File "ou_dedetai/installer.py", line 342, in ensure_launcher_shortcuts
  File "ou_dedetai/installer.py", line 320, in ensure_launcher_executable
  File "ou_dedetai/installer.py", line 311, in ensure_config_file
  File "ou_dedetai/installer.py", line 281, in ensure_product_installed
  File "ou_dedetai/installer.py", line 269, in ensure_icu_data_files
  File "ou_dedetai/installer.py", line 216, in ensure_winetricks_applied
  File "ou_dedetai/installer.py", line 190, in ensure_wineprefix_init
  File "ou_dedetai/installer.py", line 166, in ensure_product_installer_download
  File "ou_dedetai/installer.py", line 137, in ensure_premade_winebottle_download
  File "ou_dedetai/installer.py", line 131, in ensure_winetricks_executable
[...]

Issue #237, opened 2024-11-30
Change default download location to XDG_CACHE_HOME rather than downloads


ctrlaltf24:
Downloads should continue to be scanned after cache home (which is normally ~/.cache)

This is a nice to have if you like to keep your downloads folder clean

> There is a single base directory relative to which user-specific non-essential (cached) data should be written. This directory is defined by the environment variable $XDG_CACHE_HOME.
	•	https://specifications.freedesktop.org/basedir-spec/latest/

https://specifications.freedesktop.org/basedir-spec/latest/

ctrlaltf24:
It should be noted arch (which we support) doesn't create the Downloads directory by default, nor does it set XDG_DOWNLOAD_DIR.

n8marti:
Ubuntu doesn't set `$XDG_CACHE_HOME`:
```
$ env | grep -i xdg
XDG_CONFIG_DIRS=/etc/xdg/xdg-ubuntu:/etc/xdg
XDG_MENU_PREFIX=gnome-
XDG_SESSION_DESKTOP=ubuntu
XDG_SESSION_TYPE=wayland
XDG_CURRENT_DESKTOP=ubuntu:GNOME
XDG_SESSION_CLASS=user
XDG_RUNTIME_DIR=/run/user/1000
XDG_DATA_DIRS=/usr/share/ubuntu:/usr/share/gnome:/home/nate/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop
```
But we could fall back to `$HOME/.cache`.

The historical reason for using `~/Downloads` (from my perspective, anyway), is that it has allowed a naïve user to pre-download the necessary parts using their browser at a time and place where the internet is more performant, then later perform the installation with a more limited (in speed or affordability) connection. This was more useful when the only external download was the Logos offline installer. Now we have that, plus the appimage, plus ICU files, and who knows what else down the road. So I agree that it makes good sense to check both locations for the necessary files, then, if _we_ download them, put them in `$HOME/.cache` instead of `~/Downloads`.

n8marti:
I assume this would actually place files into `$HOME/.config/FaithLife-Community`, right?

ctrlaltf24:
> I assume this would actually place files into `$HOME/.cache/FaithLife-Community`, right?

Correct, we'd want to put it in there. Logic something like:
```python
[...]

Issue #236, opened 2024-11-26
Arch's dependencies not properly handled by LLI binary


thw26:
Split from #143.

Issue #143, opened 2024-07-30
Fedora's dependencies not properly handled by LLI binary


n8marti:
Fedora's `dnf` (and `yum`) is actually a python script. Since LLI includes its own python libraries, there seems to be some kind of library mismatch when the binary tries to run `dnf` commands. The only known output at the moment is that the `dnf list installed` command exits with a non-zero exit code (1).

Proposed workaround: use `rpm` to check installed packages, then provide a command to the user to run manually to install any missing dependencies.

thw26:
Another option for our Fedora users would be to host a bash script that we then execute on Fedora. They would still get automated installs that way.

thw26:
We've opted for users having to perform a manual install until we can figure out a proper fix.

thw26:
Added Arch to the issue description. We can separate these into two issues if desired.

abbasou:
> We've opted for users having to perform a manual install of dependencies with a continue question to prompt the user before proceeding until we can figure out a proper fix.

Using KDE on Fedora 40 here. When running LLI in CLI, I get `2024-08-29 13:53:07 ERROR: Your distro requires manual dependency installation.`  but no prompt to install dependencies or which dependencies are needed. I checked [here](https://github.com/FaithLife-Community/LogosLinuxInstaller/wiki/System-Dependencies) and have verified that all the packages listed are installed. Is that list up-to-date?

thw26:
Most likely we need to implement a bypass for Fedora. Most likely we are just grabbing the distro and saying no, even if all the packages are fine. Should be an easy fix.

n8marti:
If you use the most recent release (alpha-14) simply run it with `./LogosLinuxInstaller` on fedora 40 you should get something like this, which will tell you the command to run to install the dependencies:

![VirtualBox_LLI Fedora 40_29_08_2024_18_43_31](https://github.com/user-attachments/assets/3dc35cbe-3b4e-4e08-a913-d8f3cb564565)

I just did this in a virtual machine, then chose "Continue", then the TUI exited prematurely. When running it again, it no longer gives the info on how to install the dependencies, but then it *does* get stuck, as you say:

![VirtualBox_LLI Fedora 40_29_08_2024_18_50_05](https://github.com/user-attachments/assets/420b61d3-70d3-433c-8c4b-06ec654c3af1)

This is definitely a bug on our end, and we'll see if we can sort it out. I'll try to upload the log file next.

[...]

PR #293, opened 2025-02-05
fix: remove pyinstaller from LD_LIBRARY_PATH when shelling out

ctrlaltf24:
Fixes: #292 #143 #236

Didn't test:
- alpine

Tested:
- user that was able to reproduce #292 tested and it works for them
- running integration.run_cmd(["env") from a pyinstaller binary (with and with LD_LIBRARY_PATH set), observed LD_LIBRARY_PATH did not include tmpdir
- running system.run_command(["env") from a pyinstaller binary (with and with LD_LIBRARY_PATH set), observed LD_LIBRARY_PATH did not include tmpdir
- running system.run_command(["env") from a pyinstaller binary (without LD_LIBRARY_PATH set), observed LD_LIBRARY_PATH was not set
- clean install & run - login page showed successfully (install procedure calls popen_command, and run_command)
- built pyinstaller binary and ran on fedora 40 doing a clean install via TUI. login screen shown!
- built pyinstaller binary and ran on arch doing a clean install via TUI, after creating a Downloads folder (which isn't created by default on arch (tracking that [here](https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/237#issuecomment-2638329619))). login screen shown!

### DIFF
--- a/ou_dedetai/config.py
+++ b/ou_dedetai/config.py
@@ -233,6 +233,9 @@ class EphemeralConfiguration:
     
     Accepted values: tk (GUI), curses (TUI), cli (CLI)"""
 
+    wine_args: Optional[list[str]] = None
+    """Arguments when running wine from cli"""
+
     terminal_app_prefer_dialog: Optional[bool] = None
 
     # Start of values just set via cli arg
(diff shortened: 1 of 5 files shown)


n8marti (approved):
Everything looks good to me, and since you've already tested it, let's go ahead an merge it in.

Issue #278, opened 2025-01-26
Tui color scheme change causes hang


jg00dman:
Choosing to change the color scheme causes the spinning box of doom. Ou Dedetai 4.0.0-beta.6 (stable)

thw26:
cf. https://github.com/FaithLife-Community/LogosLinuxInstaller/issues/259#issuecomment-2609802552

thw26:
See @ctrlaltf24's comments in https://github.com/FaithLife-Community/LogosLinuxInstaller/pull/297 for the race condition.

PR #297, opened 2025-02-06
fix: fix setting tui color scheme

### LINKED ISSUE #259: TUI Color Scheme Not Being Saved Following Refactor

thw26:
Fixes #259.

There were two issues:

- the legacy config value was wiping out the new config value
- the config's init function, in the call to the legacy config, was setting a default value, thus overriding reading the config

To fix these:

- I changed the tui color scheme variable name to differentiate it
- I removed the setting of the default value
- I checked for a None value, which is coming from the legacy config

---

To remove the None check, a default value needs to be established, but setting the default value in the legacy config is causing an existing config's set color scheme to be ignored in favor of the default.

---

This PR also fixes the Logos color scheme: its value was improperly set.

This PR also adds the System color scheme.

---

This needs further work though as sometimes the event is not cleared, causing #278.

### DIFF
--- a/ou_dedetai/config.py
+++ b/ou_dedetai/config.py
@@ -325,7 +325,7 @@ class PersistentConfiguration:
     wine_binary_code: Optional[str] = None
     backup_dir: Optional[str] = None
 
-    # Color to use in curses. Either "Logos", "Light", or "Dark"
+    # Color to use in curses. Either "System", "Logos", "Light", or "Dark"
     curses_color_scheme: str = "Logos"
     # Faithlife's release channel. Either "stable" or "beta"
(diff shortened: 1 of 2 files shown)


ctrlaltf24 (approved):
Changed how curses_color_scheme is stored, now it handles None as a valid type to be stored, but the getter itself ensures the default value. Updated type hints appropriately

As for the hang, it looks like a race condition between:
- go_to_title_screen
- config update hook (noticing that the color changed)
- display thread

It appears as if the color is happening first, and then not properly switching off the page. When I attached a debugger, sometimes it'd work, sometimes it'd hang on the page - a tell-tale sign of a race condition

Issue #239, opened 2024-12-05
Add Snap and Flatpak Support


thw26:

ctrlaltf24:
not sure what the status of the Snap is, but it's here: https://snapcraft.io/logos10-unofficial maintained by @n8marti

PR #298, opened 2025-02-06
feat: build snap package

### LINKED ISSUE #237: Change default download location to XDG_CACHE_HOME rather than downloads

n8marti:
Also fix #237 #239
- add `constants.RUNMODE`
- default downloads to `XDG_CACHE_HOME/FaithLife-Community` 
- add `app.conf.user_download_dir` (there's a typo in the commit message...)


ou_dedetai/installer.py
```diff
@@ -349,8 +347,27 @@ def create_wine_appimage_symlinks(app: App):
...
+Comment=Manages FaithLife Bible Software via Wine
+Exec={exec_cmd}
+Icon={icon_path}
+Terminal=false
+Type=Application
+StartupWMClass={wm_class}
+Categories=Education;
+Keywords=Logos;Verbum;Bible;Control;
+"""
+    xdg_data_home = Path(os.getenv('XDG_DATA_HOME', Path.home()))
```
ctrlaltf24, line 369: Shouldn't this default to ~/.local/share?

ou_dedetai/installer.py
```diff
@@ -349,8 +347,27 @@ def create_wine_appimage_symlinks(app: App):
...
+    launcher_path = xdg_data_home / 'applications' / filename
```
ctrlaltf24, line 370: Do we need to create this directory?

ou_dedetai/utils.py
```diff
@@ -85,9 +85,10 @@ def clean_all():
 
 def get_user_downloads_dir() -> str:
     home = Path.home()
+    downloads_path = str(home) # no guarantee that ~/Downloads exists (e.g. arch)
```
ctrlaltf24, line 88: Let's revert this change, and handle this in another PR.

snap/gui/oudedetai.verbum.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
+Name=Verbum Bible
+GenericName=FaithLife Wine App Installer
+Comment=Manages FaithLife Bible Software via Wine (snap)
+Exec=oudedetai.verbum
```
ctrlaltf24, line 5: Where is this stored?

Assuming this is this the verbium app in snap.yml?

snap/gui/oudedetai.logos.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
+Name=Logos Bible
+GenericName=FaithLife Wine App Installer
+Comment=Manages FaithLife Bible Software via Wine (snap)
+Exec=oudedetai.logos
```
ctrlaltf24, line 5: Where is this stored?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,139 @@
...
+    stage-packages:
+      - libavcodec60
+      - libavformat60
+      - libavutil58
+      - libusb-1.0-0
+      - libxkbregistry0
+
+  oudedetai:
+    source: https://github.com/FaithLife-Community/LogosLinuxInstaller.git
+    # FIXME: just for testing
```
ctrlaltf24, line 105: ^ before merge

ou_dedetai/network.py
```diff
@@ -305,8 +305,7 @@ def logos_reuse_download(
     status_messages: bool = True
 ):
     dirs = [
-        app.conf.install_dir,
```
ctrlaltf24, line 308: Why was install_dir removed? It's often there

ou_dedetai/constants.py
```diff
@@ -31,11 +41,21 @@
 CACHE_LIFETIME_HOURS = 12
 """How long to wait before considering our version cache invalid"""
 
+if RUNMODE == 'snap':
+    cache_dir = Path(os.getenv('SNAP_USER_COMMON')) / '.cache'
+else:
+    try:
+        cache_dir = Path(os.getenv('XDG_CACHE_HOME'))
```
ctrlaltf24, line 48: Can this use getenv's default argument instead of a try/catch?

ou_dedetai/utils.py
n8marti: I don't mind doing that, but is it really necessary to split it out? The function still returns the user's `~/Downloads` folder if it's listed in `$XDG_CONFIG_HOME/user-dirs.dirs` or `$HOME/.config/user-dirs.dirs`. It's just the last-resort fallback that has changed.

ctrlaltf24: Idk about all users, but I take great care in organizing my top level home directory. I'd much rather even loose these caches files on reboot than have them polite my home-directory

ou_dedetai/installer.py
n8marti: oh, yes. Good catch.

ou_dedetai/installer.py
n8marti: we do below on L375 :wink:

ou_dedetai/utils.py
n8marti: We could fallback to `/tmp`, I suppose. But this function is now only defining the user's folder to *search* before initiating a new download. If we decide to initiate the download, then it will go to the new cache location instead of `~/Downloads` or `~`; i.e. (`config.py` L999+):
```python
    @property
    def download_dir(self) -> str:
        if self._download_dir is None:
            self._download_dir = constants.DEFAULT_CACHE_DIR
        return self._download_dir
```

snap/gui/oudedetai.verbum.desktop
n8marti: Yes, `snapd` doesn't recoginize new desktop files placed inside of the snap app's user's `.local/share/applications` folder (e.g. `~/snap/oudedetai/current/.local/share/applications`), so the desktop files need to be installed when the snap package is installed.

The `snap/snapcraft.yaml` then included the normal `oudedetai` command, plus two subcommands: `oudedetai.logos` and `oudedetai.verbum` that these desktop files will run. Those subcommands then run `run.sh` found in `snap/bin`, which runs the installed Faithlife app. If the app isn't installed, it first runs the `oudedetai --install-app` command.

snap/gui/oudedetai.logos.desktop
n8marti: same reply as for `oudedetai.verbum.desktop` above.

ou_dedetai/network.py
n8marti: As I mentioned in the dev chat, the only file that we download that's ever stored in `install_dir` is the `oudedetai` binary. But it's a relatively small file, and if we're trying to download that, then we are surely doing an update and we don't want to fall back on the existing (and therefore older) version of the binary. So I don't see any reason to include this folder in the existing downloads search.

ou_dedetai/utils.py
ctrlaltf24: Would rather revert then deal with this later

ou_dedetai/constants.py
ctrlaltf24: I was thinking:
```
cache_dir = Path(os.getenv("XDG_CACHE_DIR", Path.home() / ".cache")
```

n8marti: of course. Right you are.

ctrlaltf24 (approved):
LGTM, just change the branch in snapcraft.yml when you're ready to merge

PR #302, opened 2025-02-06
fix: TUI color scheme cycle hang

ctrlaltf24:
Fix: #278

remove unneeded TUI queues
also removed a clear/refresh in update_main_window_contents that may been the race condition we were looking for. This is done in draw. It renders correctly

Tested:
- Clean install of Logos and run
- changing color scheme over and over again
- navigating into utilities menu and back again
- resizing screen super small (both horizontal and vertical), screen to small was shown

### DIFF
--- a/ou_dedetai/logos.py
+++ b/ou_dedetai/logos.py
@@ -131,11 +131,8 @@ class LogosManager:
             if reason is not None:
                 logging.debug(f"Warning: Wine Check: {reason}")
             wine.wineserver_kill(self.app)
-            app = self.app
-            from ou_dedetai.gui_app import GuiApp
-            if not isinstance(self.app, GuiApp):
-                # Don't send "Running" message to GUI b/c it never clears.
-                app.status(f"Running {self.app.conf.faithlife_product}…")
+            # Don't send "Running" message to GUI b/c it never clears.
+            logging.info(f"Running {self.app.conf.faithlife_product}…")
             self.app.start_thread(run_logos, daemon_bool=False)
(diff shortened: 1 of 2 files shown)


thw26 (changes requested):
One of the changes has made the TUI color scheme selection work only once and then fail to work again. This kind of issue usually occurs because of the way choices are handled in relation to screens.

Once a choice is selected for a screen, that choice is "locked" to that instance of a screen in the current code.

Most likely what is happening is that we are interacting with an old instance of the main menu rather than the new instance, likely due to removing the switch_q.

thw26:
I just tested this again and I cannot replicate the error I just posted about. I'm unsure why it happened then but not now.

thw26:
Current replication of problem:

- Hit change color screen
- Then hit exit
- Exit will fail.
- Then after hitting Exit, changing color scheme will fail.

thw26:
This matches up with my description of the problem. So long as we keep hitting change color scheme, the choice associated with the menu remains the same. On attempting to change the choice associated with the menu, it fails, and subsequent choices then fail.

ctrlaltf24:
> This matches up with my description of the problem. So long as we keep hitting change color scheme, the choice associated with the menu remains the same. On attempting to change the choice associated with the menu, it fails, and subsequent choices then fail.

Found the bug and patched

thw26:
We will probably need to make the same change for all menus. I'm not sure when I will get to testing, but we'll need to test the installer and other menus to see if they, too, are responsive after an initial choice.

thw26 (approved):

PR #306, opened 2025-02-10
feat: add nix-shell

ctrlaltf24:
This python package list should not be considered up-to-date, see pyproject.toml for the latest.

Sets all envs needed for running on NixOS
Also fixed bug where WINE_EXE wasn't respected

on nix simply run `nix-shell` and it will download the python dependencies as well as wine needed for Logos on NixOS

Tested:
- Install on NixOS (TUI)
- Noticed that the TUI "Run Logos" button didn't work the first time, had to re-launch OD
- Noticed after logging in, when hitting "Continue" after selecting Download All Resources, Logos crashed, but recovered after launching again

Didn't see any indication the above bugs were NixOS specific and they were easily recoverable.

### DIFF
--- a/ou_dedetai/config.py
+++ b/ou_dedetai/config.py
@@ -624,9 +624,10 @@ class Config:
             # Install Dir has the name of the product and it's version. Reset it too
             if self._overrides.install_dir is None:
                 self._raw.install_dir = None
-            # Wine is dependent on the product/version selected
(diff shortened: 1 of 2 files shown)


n8marti (approved):
I assume the nix config functions as intended, since I'm not able to test, but the code looks good.

PR #309, opened 2025-02-13
feat: support flatpak

ctrlaltf24:
Also supports XDG_STATE_HOME and XDG_CONFIG_HOME

Based on wine compiled and built here: https://github.com/flathub/org.winehq.Wine it should be noted they update to latest stable wine at their digression. Presently that works with Ou Dedetai, we support wine 10. If sometime in the future they no longer have a compatible version, we can release our own/bundle it here.

Building:
```
pip install .
./scripts/build_binary.sh
BUILD_DIR=`mktemp -d`
# This downloads the dependencies
flatpak install org.winehq.Wine/x86_64/stable-24.08
flatpak-builder --install $BUILD_DIR io.github.FaithlifeCommunity.OuDedetai.yml --force-clean
# Now either use the desktop files or launch directly
flatpak run io.github.FaithlifeCommunity.OuDedetai
```

Tested:
- `flatpak run io.github.FaithlifeCommunity.OuDedetai` -> shows GUI
- `flatpak run --command=/app/bin/entrypoint.product.sh io.github.FaithlifeCommunity.OuDedetai Logos` -> Logged in, resources downloaded, Logos dashboard shown
- `cat /var/lib/flatpak/exports/share/applications/io.github.FaithlifeCommunity.OuDedetai.Logos.desktop`, then running the Exec line manually -> Logged in, resources downloaded, Logos dashboard shown

Didn't test:
- launching the .desktop from an interactive environment/some app drawer

Ideally would have some improvement:
- Right now if the user launches the desktop file from their Desktop Environment, there is no feedback that we're installing. The user may hit it again and cause issues. We could just show our GUI, but I feel like users who want a flatpak if they want to launch Logos, it should open logos and do the steps needed to do so (in this case download). Would be nice to have some user-feedback that we're downloading (especially if it goes wrong)

Fixes: #307


ou_dedetai/installer.py
```diff
@@ -397,9 +396,7 @@ def create_launcher_shortcuts(app: App):
 
     if constants.RUNMODE == 'binary':
         lli_executable = f"{installdir}/{constants.BINARY_NAME}"
-    elif constants.RUNMODE == 'snap':
-        lli_executable = f"{os.getenv('SNAP')}/bin/{constants.BINARY_NAME}"
```
ctrlaltf24, line 401: This code path wasn't reachable, as this function is only called if RUNMODE is binary. We don't really want to do this otherwise, as we want to be running in a snap

ou_dedetai/constants.py
```diff
@@ -16,13 +23,15 @@ def get_runmode() -> str:
...
+if hasattr(sys, '_MEIPASS') and (RUNMODE == 'binary' or RUNMODE == "flatpak"):
     BUNDLE_DIR = Path(sys._MEIPASS)
-elif RUNMODE == 'snap':
-    BUNDLE_DIR = Path(os.environ.get('SNAP'))
+elif _snap is not None and RUNMODE == 'snap':
+    BUNDLE_DIR = Path(_snap)
 else:
     # We are running in normal development mode
     BUNDLE_DIR = Path(__file__).resolve().parent
+del(_snap)
```
ctrlaltf24, line 34: these changes are so mypy recognizes SNAP env can't be none in this code path. We could/should? remove the RUNMODE checks, as we're duplicating them here

ou_dedetai/cli.py
```diff
@@ -63,7 +63,7 @@ def run_installed_app(self):
         self.logos.start()
         # Keep the process running so that our background threads can keep running
         while self.logos.logos_state != LogosRunningState.STOPPED:
-            time.sleep(10)
+            time.sleep(3)
```
ctrlaltf24, line 66: Lowered timeout due to initial crash when downloading resources in flatpak. When this is here, the transition is less clunky. The user is slightly confused for a second why the window just closed, but then shortly after it's open again.

Not sure why it crashes, didn't spend the time to investigate, could be a long rabbit trail.

ou_dedetai.spec
```diff
@@ -6,8 +6,8 @@ a = Analysis(
     pathex=[],
     #binaries=[('/usr/bin/tclsh8.6', '.')],
     binaries=[],
-    datas=[('ou_dedetai/img/*icon.png', 'img'),('ou_dedetai/assets/LogosStubFailOK.mst', 'assets')],
-    hiddenimports=[],
+    datas=[('ou_dedetai/img', 'img'),('ou_dedetai/assets', 'assets')],
```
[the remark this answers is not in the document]
thw26: I like it

ou_dedetai.spec
```diff
@@ -6,8 +6,8 @@ a = Analysis(
...
+    hiddenimports=["tkinter"],
```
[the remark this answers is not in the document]
thw26: I've seen this with a number of distros—they don't seem to build Python with it by default. Unsure

ou_dedetai/cli.py
thw26: This will certainly feel more responsive

ou_dedetai/constants.py
thw26: If it's duplicated, let's simplify

snap/gui/oudedetai.desktop
```diff
@@ -7,5 +7,5 @@ Icon=${SNAP}/meta/gui/oudedetai.png
 Terminal=false
 Type=Application
 StartupWMClass=Ou Dedetai
-Categories=Education;
+Categories=Education;Spirituality;
 Keywords=Logos;Verbum;Bible;Control;
```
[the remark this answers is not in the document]
thw26: Might want to add Christianity as well, here and above.

And Jesus! :D

snap/gui/oudedetai.logos.desktop
```diff
@@ -7,5 +7,5 @@ Icon=${SNAP}/meta/gui/logos.png
 Terminal=false
 Type=Application
 StartupWMClass=logos.exe
-Categories=Education;
+Categories=Education;Spirituality;
```
thw26, line 10: See above

thw26:
>Do we need to plaster unofficial over this, or some other means? Right now the flatpak's Logos .desktop could be mistaken for first party

Yes

Before testing, my only other wonder is if the flatpak env plays nicely with our wine binary finder. You say it's working, so it must!

io.github.FaithlifeCommunity.OuDedetai.yml
```diff
@@ -0,0 +1,184 @@
...
+    sources:
+      - type: inline
+        dest-filename: ld.so.conf
+        contents: |
+          /app/lib32
+          /app/lib/i386-linux-gnu
+  - name: oudedetai
+    buildsystem: simple
+    build-commands:
+      # Our installer expects sudo to be installed. Make it happy.
```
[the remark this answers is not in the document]
ctrlaltf24: Surprisingly not, wine comes from the base flatpak, we don't use much else, most of the other stuff is bundled in the binary

io.github.FaithlifeCommunity.OuDedetai.yml
```diff
@@ -0,0 +1,184 @@
...
+          [Desktop Entry]
+          Name=Ou Dedetai
+          GenericName=FaithLife Wine App Installer
+          Comment=Manages FaithLife Bible Software via Wine
+          Exec=/app/bin/entrypoint.sh
+          Icon=/app/share/icons/hicolor/512x512/apps/oudedetai.png
+          Terminal=false
+          Type=Application
+          StartupWMClass=Ou Dedetai
+          Categories=Education;Spirituality;
```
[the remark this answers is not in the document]
ctrlaltf24: Sounds good. Languages (plural) (https://specifications.freedesktop.org/menu-spec/latest/additional-category-registry.html)

thw26:
Tested. Works. Approved once changes above are addressed.

thw26 (approved):

PR #321, opened 2025-02-20
fix: Alpine installs

### LINKED ISSUE #294: Alpine Installs No Longer Work
It looks like beta 6 broke Alpine installs. #293 fixes portions of Alpine's install, but more changes are needed.

thw26:
Fixes #294.

Work in Progress.

The issue here appears to be that the Wine AppImage was not made with muslc, and so thus it won't run even if everything else is up and running.

Now when running the appimage, it says it can't open the wine commands. I believe that is for the above reason.

### DIFF
--- a/ou_dedetai/system.py
+++ b/ou_dedetai/system.py
@@ -482,7 +482,7 @@ def get_package_manager() -> PackageManager | None:
         packages = (
             "bash bash-completion "  # bash support
             "gcompat "  # musl to glibc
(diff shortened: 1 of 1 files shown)


ctrlaltf24 (approved):
I'm fine with disabling appimage on alpine or recommending flatpak

thw26:
Made changes to remove appimages from the installer for Alpine. May need to be expanded in the future for postmarketOS. We may need to consider lumping distros into an array.

PR #322, opened 2025-02-20
chore: remove Logos 9 support

### LINKED ISSUE #53: Remove Logos 9 Support
After posting a discussion and calling the community to respond in Telegram/Matrix, after being open for three weeks we have not had any response pertaining to keeping Logos 9 support.

https://github.com/FaithLife-Community/LogosLinuxInstaller/discussions/44

I suggest that we remove Logos 9 support. We can keep most of the infrastructure we have in place should a Logos 11 be forthcoming and we need to manage both versions. This would obviate the need to have a version selection menu/dropdown.

In general, I am in favor of supporting only the latest release of Logos, given that it is proprietary software. Of course #22 is an exception to this rule.

thw26:
Fixes #53.

Removes Logos 9 support. Since the FLPRODUCT version in @ctrlaltf24's abstract class refactor is now acquired through the config, all we need to do is have it return 10 no matter what.

I've left the question in code in order to maintain should we need it in the future. I don't think we will since Logos has moved towards a semi-SaaS model.

### DIFF
--- a/ou_dedetai/config.py
+++ b/ou_dedetai/config.py
@@ -612,9 +612,11 @@ class Config:
     def faithlife_product_version(self) -> str:
         if self._overrides.faithlife_product_version is not None:
             return self._overrides.faithlife_product_version
-        question = f"Which version of {self.faithlife_product} should the script install?: "  # noqa: E501
-        options = constants.FAITHLIFE_PRODUCT_VERSIONS
-        return self._ask_if_not_found("faithlife_product_version", question, options, []) # noqa: E501
+        if self.conf.faithlife_product_version is None:
+            return "10"  # Keep following for historical reasons.
+        #question = f"Which version of {self.faithlife_product} should the script install?: "  # noqa: E501
+        #options = constants.FAITHLIFE_PRODUCT_VERSIONS
+        #return self._ask_if_not_found("faithlife_product_version", question, options, []) # noqa: E501
 
     @faithlife_product_version.setter
     def faithlife_product_version(self, value: Optional[str]):
--- a/ou_dedetai/utils.py
+++ b/ou_dedetai/utils.py
@@ -122,12 +122,10 @@ def install_dependencies(app: App):
     if targetversion == 10:
         system.install_dependencies(app, target_version=10)  # noqa: E501
     elif targetversion == 9:
-        system.install_dependencies(
-            app,
-            target_version=9
-        )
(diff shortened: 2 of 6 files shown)


ou_dedetai/config.py
```diff
@@ -612,9 +612,10 @@ def faithlife_product(self, value: Optional[str]):
     def faithlife_product_version(self) -> str:
         if self._overrides.faithlife_product_version is not None:
             return self._overrides.faithlife_product_version
-        question = f"Which version of {self.faithlife_product} should the script install?: "  # noqa: E501
-        options = constants.FAITHLIFE_PRODUCT_VERSIONS
-        return self._ask_if_not_found("faithlife_product_version", question, options, []) # noqa: E501
+        return "10"  # Bypass the check. Keep the rest for historical reasons.
```
ctrlaltf24, line 615: This will overnight the config on next write. Instead let's check to make sure the config value is unset before forcing it to 10

The reason I'm suggesting this is in the future we may want to run part of the installation process on upgrade of OD, and we should know whether or not to skip (I.e if it's logos 9)

ou_dedetai/utils.py
```diff
@@ -121,13 +121,8 @@ def install_dependencies(app: App):
 
     if targetversion == 10:
         system.install_dependencies(app, target_version=10)  # noqa: E501
-    elif targetversion == 9:
```
ctrlaltf24, line 124: Suggestion: keep the elif but exit the app if target version is 9 stating we no longer support it

In case they set the target version env

ou_dedetai/config.py
thw26: See latest force if sufficient.

ou_dedetai/utils.py
thw26: See latest force if sufficient.

ou_dedetai/utils.py
```diff
@@ -122,12 +122,10 @@ def install_dependencies(app: App):
     if targetversion == 10:
         system.install_dependencies(app, target_version=10)  # noqa: E501
     elif targetversion == 9:
-        system.install_dependencies(
-            app,
-            target_version=9
-        )
+        app.status("Logos 9 not supported", 100)
```
ctrlaltf24, line 125: suggestion: app.exit() it calls sys.exit

ctrlaltf24 (approved):

PR #324, opened 2025-02-20
Add TUI installer menu

thw26:
In mirroring the workflow of the GUI, this now implements the option for a default install where the user only selects the product and a custom install, where the user can change the install variables.

The initial implementation was found to error during install if the user hit return to main menu. This required refactoring the code to avoid in a few different ways.


ou_dedetai/config.py
```diff
@@ -708,7 +708,8 @@ def install_dir(self) -> str:
     
     @install_dir.setter
     def install_dir(self, value: str | Path):
-        value = str(Path(value).absolute())
+        if value is not None:
```
thw26, line 711: Without this, the variable reset in the installer menu errors.

ctrlaltf24: Update the type hint to Optional[str | Path]

ou_dedetai/tui_app.py
```diff
@@ -481,7 +483,7 @@ def choice_processor(self, stdscr, screen_id, choice):
         }
 
         # Capture menu exiting before processing in the rest of the handler
-        if screen_id not in [0, 2] and (choice in ["Return to Main Menu", "Exit"]):
+        if screen_id not in [0] and (choice in ["Return to Main Menu", "Exit"]):
```
ctrlaltf24, line 486: Why was 2 removed from here? That's handle ask response

ctrlaltf24 (commented):
Mirroring GUI shouldn't install or custom be options on the main page? Prevents an unnecessary user input


What's this about going to main menu? Does the current code have this failing? IMO the TUI should exit if an install is aborted (by going to the main menu) - more complicated to try to cleanup

ou_dedetai/tui_app.py
thw26: So this was removed because all/most of the questions are handle ask response and are thus being recorded when the user chooses them. When removed, I am able to reset the custom installer in case the user makes a mistake (e.g., selected Logos instead of Verbum). Without this change, the user would have to delete their config to correct this, which I think is bad UX.

thw26:
In terms of whether the pre-install main menu should have one option (Install → default or custom) or two (install default or install custom), either option can work. On one level I like having hierarchical options gathered, on another level I like flat options.

The code is robust enough to handle either. The question is which is best. I think the answer to that fluctuates in the software community. Which do we prefer? Hierarchical or flat?

My folder structure for my personal documents is "lightly hierarchical" structure that aims to be as flat as possible, while still allowing directory use, as but a personal example.

These options disappear once we have an install, so would we rather have one install option with choices, like the Logos installer offering typical or custom installs, or do we want each command immediately user facing?

I am six of one, half a dozen of another.

ou_dedetai/tui_app.py
ctrlaltf24: Perhaps we shouldn't have an option in ask to return to the main menu at all, perhaps that should be an Exit

n8marti:
> In terms of whether the pre-install main menu should have one option (Install → default or custom) or two (install default or install custom), either option can work. On one level I like having hierarchical options gathered, on another level I like flat options.

I vote for two up-front options. This lets the most-common use case happen with one choice (with the side effect that the other use case *also* happens with one choice). But that's the test I tend to apply to UI decisions: Which option streamlines the user experience for *most* users? Advanced users are more than happy to click through menus in order to have their tailored experience. "Regular users" would rather the thing just be done automatically, if possible. As an example of a UI that doesn't seem to follow this philosophy, see VLC. I think it's a terrible UI for "normal" people.

ou_dedetai/tui_app.py
thw26: I think we should maintain the return to main menu option. I think being in the app should return you to the app, or offer a way to fix the issue. Put another way, I don't think we should associate the lifetime of the installer with the lifetime of the app.

thw26: Perhaps we could consider this as two issues:

- Should we have a return to main menu option
- Should the installer reset the config on a failed or exited install, so that the user can change a choice in the event of a mistake.

thw26:
Latest force flattens the menu structure

ou_dedetai/tui_app.py
```diff
@@ -519,30 +520,51 @@ def go_to_main_menu(self):
...
         if choice is None or choice == "Exit":
             logging.info("Exiting installation.")
             self.tui_screens = []
             self.llirunning = False
-        elif choice.startswith("Install"):
-            self.reset_screen()
-            self.installer_step = 0
-            self.installer_step_count = 0
+
+        if choice.endswith("Install"):
```
ctrlaltf24, line 537: nit: would prefer precise string matching

if choice in ["Default Install", "Advanced Install"]

ou_dedetai/tui_app.py
```diff
@@ -519,30 +520,51 @@ def go_to_main_menu(self):
     def main_menu_select(self, choice):
         def _install():
             try:
+                if self._cancel_install_event.is_set():
+                    return
                 installer.install(app=self)
+                if self._cancel_install_event.is_set():
```
ctrlaltf24, line 526: Does this work? I wouldn't expect it too, since the majority of the execution is in the install function, it's unlikely that setting this will actually stop the thread (since installer.install is still running)

ou_dedetai/tui_app.py
ctrlaltf24: Ideally yes separate lifetimes, but that's far more complicated to maintain. If the user is .ask'd a question, some thread somewhere will be hung waiting for a response. It's not reasonable to recover easily in this state, and it's easy enough to restart the app (up arrow + enter) if the user wants to abort the install and try again

ctrlaltf24 (commented):
Not convinced the cancel install event actually does anything, not near computer to test

ou_dedetai/tui_app.py
```diff
@@ -894,6 +916,20 @@ def set_tui_menu_options(self):
 
         return options
 
+    def set_installer_choice_menu_options(self):
```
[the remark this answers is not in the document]
thw26: Yes, I thought I removed it

ou_dedetai/tui_app.py
thw26: Can do. This depends on the renaming above.

ou_dedetai/tui_app.py
thw26: I'll need to do more thorough testing to verify.

ou_dedetai/tui_app.py
thw26: Could we perhaps run some kind of config clean up on a botched install to reset the config vars?

ou_dedetai/tui_app.py
thw26: The testing I did before submitting did make this work, though I imagine there will be threading issues.

ou_dedetai/tui_app.py
```diff
@@ -879,7 +901,7 @@ def set_tui_menu_options(self):
                 indexing = "Run Indexing"
             labels_default = [run, indexing]
         else:
-            labels_default = ["Install Logos Bible Software"]
+            labels_default = ["Default Install", "Advanced Install"]
```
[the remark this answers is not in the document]
ctrlaltf24: Synergy with the GUI makes sense, as stated later exact string matching is preferred (in case we add another option later that happens to end with Install)

ou_dedetai/tui_app.py
ctrlaltf24: I'm less worried about the config and more worried about the thread

thw26: Reverted the change that prompted this thread.

I will open an issue for the prior comment.

ou_dedetai/tui_app.py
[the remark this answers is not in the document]
thw26: Switched labels to match GUI.

thw26:
I forced another set of changes reverting some of those which raised questions.

PR #327, opened 2025-02-21
refactor: create portable config

thw26:
Creates a portable config under the install_dir, positioned at top for ease of finding.

Refactors the write_config function by adding a write_json_file function.


ctrlaltf24 (approved):
This work is complete, a follow on PR needs to copy this from the default install directory location (once it's static in the data directory).

PR #330, opened 2025-02-23
fix: don't clobber user's faithlife_product_version

ctrlaltf24:
Only return 10 if the user hasn't selected something else

Tested:
- set faithlife_product_version to 9 in config
- ran installer
- observed asking for logos 9 target versions
- observed "Logos 9 is not supported" then exit

Also tested:
- removed faithlife_product_version from config
- export TARGETVERSION=9
- ran installer
- observed asking for logos 9 target versions
- observed "Logos 9 is not supported" then exit


n8marti (approved):
This looks good. The part that I hadn't yet gotten my head around was how to distinguish between config already set in the config file vs. that coming from env variables. Since you've already tested it, I'll go ahead and merge it in.

Issue #241, opened 2024-12-05
Passive installs and ELUA agreement


ctrlaltf24:
When installing with the --passive flag, the user never sees the Faithlife ELUA before the product is installed. Out of respect for Faithlife we should prompt the EULA. 

Recommended implementation:
- Add another cli argument called `--i-agree-to-the-faithlife-elua` to agree non-interactively
- If the passive argument was specified (and the aforementioned flag wasn't), prompt the user to agree to the ELUA (providing a link to it)

PR #331, opened 2025-02-23
fix: ensure user agrees to the EULA even in non-interactive settings

ctrlaltf24:
Another route we could go is always doing a passive install, as now we have the proper EULA checks

Fixes: #241

Tested:

Scenario CLI passive:
- passed args --install-app --passive
- Asked on CLI to agree to the EULA
- install MSI success

Scenario CLI assume yes:
- passed args --install-app -y
- Installer exits requiring the --i-agree-to-faithlife-terms flag

Scenario CLI assume yes agreed to terms:
- passed args --install-app -y --i-agree-to-faithlife-terms
- Install MSI success

Scenario GUI passive:
- passed env DIALOG=tk and args --passive
- Asked on the GUI to agree to the EULA
- accepted
- install MSI success

Scenario GUI passive decline:
- passed env DIALOG=tk and args --passive
- Asked on the GUI to agree to the EULA
- declined
- Error dialog, then exit

Scenario TUI passive:
- passed args --passive
- Asked on the TUI to agree to the EULA
- accepted
- install MSI success

Scenario TUI passive decline:
- passed args --passive
- Asked on the TUI to agree to the EULA
- declined
- Shown status why we're exiting then closes after 30s

### DIFF
--- a/ou_dedetai/app.py
+++ b/ou_dedetai/app.py
@@ -124,10 +124,11 @@ class App(abc.ABC):
 
         return answer
 
-    def approve_or_exit(self, question: str, context: Optional[str] = None):
+    def approve_or_exit(self, question: str, context: Optional[str] = None) -> bool:
         """Asks the user a question, if they refuse, shutdown"""
         if not self.approve(question, context):
             self.exit(f"User refused the prompt: {question}")
+        return True
 
     def approve(self, question: str, context: Optional[str] = None) -> bool:
         """Asks the user a y/n question"""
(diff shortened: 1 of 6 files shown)


n8marti:
To be clear, you've only tested in CLI, but the implementation should work in any UI, making use of the UI's Ask* methods, right?

ctrlaltf24:
Correct on both accounts

ctrlaltf24:
Tested and fixed pre-existing ask/approve or exit UX issues on TUI and GUI

n8marti:
I don't see any problems with this. I'm glad you've also tested on GUI and TUI to smoke out the UX issues. I'm going to give @thw26 a chance to chime in before merging.

thw26:
Just tried running on TUI but I was not asked to accept terms.

I ran:

./script/run_app.py --passive

PR #332, opened 2025-02-23
feat: add install GUI

ctrlaltf24:
Leverages the generic nature of App to create a dialog that just does the installation

Useful in flatpak/snap if we want to start the install process and show the user feedback in a GUI fashion

Tested:
- setting DIALOG=tk
- passing arg --install-app
- observed install window opened, and install success

Also tested:
- setting DIALOG=tk
- passing neither -C or --install-app
- observed control panel opened

### DIFF
--- a/ou_dedetai/config.py
+++ b/ou_dedetai/config.py
@@ -438,7 +438,7 @@ class PersistentConfiguration:
                 if (isinstance(v, str) and v.startswith(self.install_dir)): #noqa: E501
                     output[k] = utils.get_relative_path(v, self.install_dir)
 
-            portable_config_path = os.path.expanduser(self.install_dir + f"/{constants.BINARY_NAME}.json")
+            portable_config_path = os.path.expanduser(self.install_dir + f"/{constants.BINARY_NAME}.json") #noqa: E501
             self.write_json_file(output, portable_config_path)
 
         self.write_json_file(output, config_file_path)
@@ -766,6 +766,8 @@ class Config:
         # Return the full path so we the callee doesn't need to think about it
         if self._raw.wine_binary is not None and not Path(self._raw.wine_binary).exists() and (Path(self.install_dir) / self._raw.wine_binary).exists(): # noqa: E501
             return str(Path(self.install_dir) / self._raw.wine_binary)
+        if not Path(output).exists():
+            logging.warning(f"Wine binary {output} doesn't exist")
         return output
 
     @wine_binary.setter
--- a/ou_dedetai/constants.py
+++ b/ou_dedetai/constants.py
@@ -59,7 +59,7 @@ if RUNMODE == 'snap':
 else:
     CACHE_DIR = Path(os.getenv('XDG_CACHE_HOME', Path.home() / '.cache' / 'FaithLife-Community')) #noqa: E501
 
-XDG_DATA_HOME = Path(os.getenv('XDG_DATA_HOME', Path.home() / '.local/share' / 'FaithLife-Community'))
+XDG_DATA_HOME = Path(os.getenv('XDG_DATA_HOME', Path.home() / '.local/share' / 'FaithLife-Community')) #noqa: E501
 CONFIG_DIR = os.getenv("XDG_CONFIG_HOME", "~/.config") + "/FaithLife-Community"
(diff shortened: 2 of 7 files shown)


n8marti:
I like the idea. I've been wanting to add this for the snap package, too. A couple of issues I've come across while testing:
- I think we need to set the window height for the status window to be adaptive, or minimal. I can't quickly find where we've set it, but this install window seems to have the same height as the Control Panel, but that's too tall.
- The Ask window fits the question and the drop-down (when choosing install dir) into separate columns from the buttons. I think it would look best if the drop-down spanned the width of both buttons and the question took up the rest of the columns to the left.
- In my own testing my passed `WINE_EXE=` isn't respected. It keeps trying to download the AppImage, but I'm passing the local binary specifically I don't want to download the appimage at this time (too big!).

n8marti:
Another comment that's more of a wish than a request right now: This has come up before, and I was reminded of it while testing this out, but I think it would be helpful to have a 2nd progress bar specifically for long-running sub-tasks during install. Specifically, at least the download steps (they take forever on my connection) and maybe the checksum comparisons (not nearly as slow as the downloads, though). In the case of the downloads the progress is determinate, but for the checksums it's not.

ctrlaltf24:
> Another comment that's more of a wish than a request right now: This has come up before, and I was reminded of it while testing this out, but I think it would be helpful to have a 2nd progress bar specifically for long-running sub-tasks during install. Specifically, at least the download steps (they take forever on my connection) and maybe the checksum comparisons (not nearly as slow as the downloads, though). In the case of the downloads the progress is determinate, but for the checksums it's not.

Since the abstract class is leveraged that issue can be tackled separately. The entire install on my machine takes less than a minute now thanks to no more font downloads

ctrlaltf24:
> - I think we need to set the window height for the status window to be adaptive, or minimal. I can't quickly find where we've set it, but this install window seems to have the same height as the Control Panel, but that's too tall.

Fixed 

> - The Ask window fits the question and the drop-down (when choosing install dir) into separate columns from the buttons. I think it would look best if the drop-down spanned the width of both buttons and the question took up the rest of the columns to the left.

Fixed

> - In my own testing my passed `WINE_EXE=` isn't respected. It keeps trying to download the AppImage, but I'm passing the local binary specifically because I don't want to download the appimage at this time (too big!).

Is the wine_binary_code set in your config? Maybe we shouldn't persist that value

n8marti:
The fixes look nice. Unfortunately, in the meantime wine from the ubuntu repo seems to have changed the way it works a bit, and it now no longer creates a `wine64` binary:
```
$ which wine
/usr/bin/wine
$ which wine64
$ ll $(which wine)
lrwxrwxrwx 1 root root 24 Feb 22 18:11 /usr/bin/wine -> /opt/wine-devel/bin/wine*
$ find /opt/wine-devel/bin -name 'wine*'
/opt/wine-devel/bin/winemaker
/opt/wine-devel/bin/wineboot
/opt/wine-devel/bin/wine
/opt/wine-devel/bin/winedump
/opt/wine-devel/bin/winebuild
/opt/wine-devel/bin/winegcc
/opt/wine-devel/bin/winepath
/opt/wine-devel/bin/winedbg
/opt/wine-devel/bin/wineserver
/opt/wine-devel/bin/wineconsole
/opt/wine-devel/bin/wineg++
/opt/wine-devel/bin/winecfg
/opt/wine-devel/bin/winemine
/opt/wine-devel/bin/winecpp
/opt/wine-devel/bin/winefile
$
```
So I'm no longer able to test that way, but I'll try with a different `WINE_EXE`.

n8marti:
before merging, please update `snap/gui/*.desktop` files to correctly include `DIALOG=tk` in each Exec line. You'll need to double-check, but I think the correct syntax is:
```
Exec=env DIALOG=tk oudedetai{.logos|.verbum}
```

n8marti (changes requested):
See previous comment for request changes to `.desktop` files in `snap/gui`.

n8marti (approved):
Fair enough. I'm happy with that, even though it will function differently than if someone downloads and runs the binary release directly, where running in the terminal defaults to the TUI, while double-clicking the binary runs the GUI.

Issue #333, opened 2025-02-23
On first boot the install button the GUI stays greyed out


ctrlaltf24:
Reproduction:
- no network cache
- launch 4.0.0-beta.8 with DIALOG=tk
- observe install button stays greyed out

Workaround: restart OuDedetai

jleaders:
I'm still seeing this issue on latest Linux Mint 

![Image](https://github.com/user-attachments/assets/ee085381-17f6-43c3-9c07-1fc878794a94)

PR #334, opened 2025-02-24
fix: also call config update hook when network cache gets populated

ctrlaltf24:
Fixes: #333

Tested:
- removed config and network cache
- launched GUI
- observed Install buttons start off disabled, then enable themselves


thw26 (approved):
Tested and approved.

PR #337, opened 2025-02-24
feat: add unittests

n8marti:
Run all unittests with `python -m unittests discover`, or use helper script `./scripts/run-tests.sh`.

Many tests are placeholders and/or skipped with either "TODO" messages or other messages explaining why the test seems unneeded or unhelpful.


tests/__init__.py
```diff
@@ -0,0 +1,10 @@
+import logging
+from pathlib import Path
+
+# Disable most logging for unit tests.
```
ctrlaltf24, line 4: The unit test library doesn't handle this? Guess not.

Pytest does if a test fails it shows output otherwise it swallows it by default

When running on GHA we want all output on (so we don't need to re-run locally to diagnose every failure)

tests/data/config.json
ctrlaltf24, line 1: not: rename to old config name so it's clear this is legacy

tests/data/wine_tiny_v2.AppImage
ctrlaltf24, line 1: Checking in an appimage? What for?

tests/test_config.py
```diff
@@ -0,0 +1,32 @@
...
+import shutil
+import unittest
+from pathlib import Path
+
+import ou_dedetai.config as config
+
+
+class TestConfigReadWrite(unittest.TestCase):
+    def setUp(self):
+        self.cfg = Path('tests/data/subdir/test_config.json')
```
ctrlaltf24, line 11: Should this use a tempfile?

tests/test_config.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @unittest.skip('TODO')
+    def test_update_config_file(self):
```
ctrlaltf24, line 14: Suggestion:

Perform operation with the code you want to test

Verify with a simple check (such as reading json)

ou_dedetai/constants.py
```diff
@@ -41,6 +41,7 @@ def get_runmode() -> str:
 APP_NAME = 'Ou Dedetai'
 BINARY_NAME = 'oudedetai'
 PACKAGE_NAME = 'ou_dedetai'
+REPO_NAME = 'OuDedetai'
```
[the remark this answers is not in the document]
n8marti: oh, well the repo link is. *That* should be updated to make use of the `REPO_NAME`.

tests/data/wine_tiny_v2.AppImage
n8marti: This is used to test the `utils.is_appimage` function, which inspects certain bytes of the AppImage file to see if it tryly is an AppImage, and if so, which version.

tests/test_config.py
n8marti: same comment as above: this file hasn't been updated yet for this PR

ctrlaltf24 (approved):

ou_dedetai/utils.py
```diff
@@ -576,9 +584,10 @@ def get_downloaded_file_path(download_dir: str, filename: str):
 
 def grep(regexp, filepath):
     fp = Path(filepath)
-    found = False
+    found = None
```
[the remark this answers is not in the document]
n8marti: No, good catch. I removed it.

PR #338, opened 2025-02-24
chore: build Arm64 binary

### LINKED ISSUE #238: Build an ARM64 Binary
There are two means of ARM64 support: one is running our x86_64 binary on ARM64, another is running a native ou dedetati on ARM64 and then running an ELF interpreter within it.

#119 at this moment shows a lot of promise for running our software on ARM64 through ELF interpretation. I'd like to carry this one step further by releasing ARM64 builds. As part of #119, we should build an ARM64 binary.

thw26:
Fixes #238. Adds a runner matrix to build Arm64 binaries.

The main purpose of this PR at the moment is to be better able to test what is needed for running Logos x86_64 through an ELF interpreter for Arm. At present we have to run both OD and Logos through the ELF interpreter. The goal of this is to be passing only Logos through the ELF interpreter.


ctrlaltf24 (changes requested):
Needs a build matrix to run the build twice, as-is it's telling GHA that either are acceptable runners when really we meant run the build twice, one for each 
https://www.blacksmith.sh/blog/matrix-builds-with-github-actions

thw26:
Great—thanks for confirming that for me. Do you think it will be sufficient just to copy the existing x86_64 over to a new one?

ctrlaltf24 (commented):
Is this tested?

ctrlaltf24:
> How would I test it?

You can kick off a build here: https://github.com/FaithLife-Community/OuDedetai/actions/workflows/build-branch.yml targeting the `fix-238` branch. I just started one, we'll see what happens.

ctrlaltf24 (approved):

PR #343, opened 2025-02-25
ci: add job to build snap

ctrlaltf24:
successful run here: https://github.com/ctrlaltf24/OuDedetai/actions/runs/13528272169

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -71,26 +71,6 @@ apps:
     plugs: *allplugs
 
 parts:
-  bin:
-    source: snap/bin
-    plugin: dump
-    organize:
-      run.sh: bin/run.sh
-  #   stage-packages:
-  #     - jq
-
-  assets:
-    source: ou_dedetai/assets
-    plugin: dump
-    organize:
-      '*': assets/
-
-  img:
-    source: ou_dedetai/img
-    plugin: dump
-    organize:
-      '*': img/
-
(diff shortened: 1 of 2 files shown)


n8marti:
Why not use the same git command for the snap version as in the snapcraft.io builds? Wouldn't that make it clearer when GHA builds are (or should be) identical to snapcraft.io builds?

snap/snapcraft.yaml
```diff
@@ -130,6 +110,18 @@ parts:
...
+    plugin: dump
+    organize:
+      snap/bin/run.sh: bin/run.sh
+      ou_dedetai/assets: assets/
+      ou_dedetai/img: img/
+    stage:
+      - bin/run.sh
+      - assets/
+      - img
+
```
n8marti, line 124: This is a nice simplification. If we're making these changes, might as well update the repo name to use "OuDedetai".

ctrlaltf24:
> Why not use the same git command for the snap version as in the snapcraft.io builds?

Not sure what you mean, are you talking about the difference in the git command via the sed? Those changes were made because there is no tag (nor should be) on non-release versions

thw26 (approved):

snap/snapcraft.yaml
ctrlaltf24: We can do that later, we're releasing now and it'd be nice to have this. We need to change the repo name in other places as well

Issue #86, opened 2024-03-29
ToolTip Incorrect in GUI on an Uninitiated Install


thw26:
While testing #83, I found that running the app has the wrong info in the tooltip for updating to the latest appimage.

The tooltip says that the button is disabled (not true) because the program wasn't installed using an appimage (hasn't happened yet).

We need to figure out how to differentiate between an install that has run and one that hasn't so that the tooltip can then read this properly.

Perhaps #85 will resolve this?

PR #344, opened 2025-02-25
feat: detect first time installs

ctrlaltf24:
Detect First time installs and warn the user if we're still in that state. This may be useful in debugging.

It should be noted we should never see this if there was no crash, this database flag is only set during resource download before the first run dialog is closed. The first run dialog would have had to crash in order for this to display

Opted not to manually mangle the logos database into force a non-first run state due to potential side effects that may last the entire install. It's safer to act as much like the window install as we can. If push comes to shove, we can bypass it as-needed, but it may have far-reaching side effects

### DIFF
--- a/ou_dedetai/logos.py
+++ b/ou_dedetai/logos.py
@@ -158,12 +158,10 @@ class LogosManager:
         if self.app.conf._logos_appdata_dir is None:
             return
         logos_appdata_dir = Path(self.app.conf._logos_appdata_dir)
-        # The glob here is for a user identifier
-        db_glob = './Documents/*/LocalUserPreferences/PreferencesManager.db'
-        results = list(logos_appdata_dir.glob(db_glob))
-        if not results:
+        logos_user_id = self.app.conf._logos_user_id
+        if not logos_user_id:
             return None
-        db_path = results[0]
+        db_path = logos_appdata_dir / "Documents" / logos_user_id / "LocalUserPreferences" / "PreferencesManager.db" #noqa: E501
         sql = (
             """UPDATE Preferences SET Data='<data """ +
             ('OptIn="true"' if val else 'OptIn="false"') +
@@ -178,12 +176,10 @@ class LogosManager:
         if self.app.conf._logos_appdata_dir is None:
             return
         logos_appdata_dir = Path(self.app.conf._logos_appdata_dir)
-        # The glob here is for a user identifier
-        db_glob = './Data/*/UpdateManager/Updates.db'
-        results = list(logos_appdata_dir.glob(db_glob))
-        if not results:
+        logos_user_id = self.app.conf._logos_user_id
+        if logos_user_id is None:
             return None
-        db_path = results[0]
+        db_path = logos_appdata_dir / "Data" / logos_user_id / "UpdateManager" / "Updates.db" #noqa :E501
         # FIXME: I wonder if we can use the result of these deletion using RETURNING
         # Then we could notify the user that there are updates.
         # If we do that we'd have to consider if their other resources are up to date
--- a/ou_dedetai/logos.py
+++ b/ou_dedetai/logos.py
@@ -8,6 +8,7 @@ import logging
 import psutil
 import threading
 
+from ou_dedetai import database
 from ou_dedetai.app import App
 
 from . import system
@@ -167,7 +168,7 @@ class LogosManager:
             ('OptIn="true"' if val else 'OptIn="false"') +
             """ StartDownloadHour="0" StopDownloadHour="0" MarkNewResourcesAsCloud="true" />' WHERE Type='UpdateManagerPreferences'""" #noqa: E501
         )
-        self.app.start_thread(utils.watch_db, str(db_path), [sql])
+        self.app.start_thread(database.watch_db, str(db_path), [sql])
 
     def prevent_logos_updates(self):
         """Edits Logos' internal database to remove pending installers
@@ -201,7 +202,7 @@ class LogosManager:
             "UPDATE Resources SET Status=1, UpdateId=NULL WHERE UpdateId IS NOT NULL "+
                 "AND UpdateId NOT IN (SELECT UpdateId FROM Updates)"
         ]
-        self.app.start_thread(utils.watch_db, str(db_path), sql)
+        self.app.start_thread(database.watch_db, str(db_path), sql)
 
     # Also noticed if the database Data/*/CloudResourceManager/CloudResources.db 
     # table TransitionStates has a ResourceId that isn't registered in UpdateManager,
--- a/ou_dedetai/repair.py
+++ b/ou_dedetai/repair.py
@@ -152,11 +152,15 @@ def detect_and_recover(ephemeral_config: EphemeralConfiguration):
                 f"{persistent_config.faithlife_product}?"
             )
             context=(
-                "The following recovery method is helpful if resource downloading is "
-                "crashing, but may not be required.\n"
+                "The following recovery method is not recommended unless "
+                "downloading resources is crashing "
+                "(i.e. the 'Continue' button causes a crash).\n"
                 "You will need to download your resources manually in the Library tab. "
                 "Use the filter 'Not on This Device' and use CTRL+A to "
-                "make this easier."
+                "make this easier.\n"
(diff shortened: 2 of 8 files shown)


ou_dedetai/repair.py
```diff
@@ -76,10 +116,17 @@ def detect_and_recover(ephemeral_config: EphemeralConfiguration):
...
+    # Recovery detection is best-effort.
+    # Since it runs very early in the app and may be complex, we don't want a
+    # bug here to interfere with normal operations.
+    try:
+        detected_failure = detect_broken_install(
+            logos_appdata_dir,
+            persistent_config.faithlife_product
+        )
+    except Exception:
+        logging.exception("Failed to check to see if installation is broken.")
```
thw26, line 128: We might want this to say something like, "Unable to detect an issue." We didn't fail our check in this case; we did check, but we didn't find anything.

ou_dedetai/logos.py
```diff
@@ -158,12 +158,10 @@ def set_auto_updates(self, val: bool):
...
-        # The glob here is for a user identifier
-        db_glob = './Documents/*/LocalUserPreferences/PreferencesManager.db'
-        results = list(logos_appdata_dir.glob(db_glob))
-        if not results:
+        logos_user_id = self.app.conf._logos_user_id
+        if not logos_user_id:
             return None
-        db_path = results[0]
+        db_path = logos_appdata_dir / "Documents" / logos_user_id / "LocalUserPreferences" / "PreferencesManager.db" #noqa: E501
         sql = (
```
thw26, line 165: I think we should put SQL operations into a db.py module.

ou_dedetai/utils.py
```diff
@@ -653,6 +653,14 @@ def parse_bool(string: str) -> bool:
     return string.lower() in ['true', '1', 'y', 'yes']
 
 
+def execute_sql(path: str | Path, sql_statements: list[str]) -> list[Any]:
```
thw26, line 656: As above; I think we should be SQL operations into a db.py module.

ou_dedetai/repair.py
```diff
@@ -44,6 +45,45 @@ def detect_broken_install(
...
+        # No other checks we can preform without the logos_user_id
+        return None
+
+    # Recovery is best-effort we don't want to crash the app on account of failures here
+    try:
+        local_user_prefrences_path = logos_app_dir / "Documents" / logos_user_id / "LocalUserPreferences" / "PreferencesManager.db" #noqa: E501
+        contents = ou_dedetai.utils.execute_sql(
+            local_user_prefrences_path,
+            [
+                "SELECT Data FROM Preferences WHERE `Type`='AppLocalPreferences' LIMIT 1" #noqa: E501
```
thw26, line 62: See comments on db.py module.

ou_dedetai/repair.py
```diff
@@ -98,6 +145,56 @@ def _run(app: App):
...
+        def _run(app: App):
+            question=(
+                "Do you want to skip the first run dialog and go straight into "
+                f"{persistent_config.faithlife_product}?"
+            )
+            context=(
+                "The following recovery method is not recommended unless "
+                "downloading resources is crashing "
+                "(i.e. the 'Continue' button causes a crash).\n"
+                "You will need to download your resources manually in the Library tab. "
```
thw26, line 158: Perhaps add an additional newline before this, or add an inline heading like: "Instructions: " or bullet points to make it clear we are giving directions.

ou_dedetai/repair.py
```diff
@@ -98,6 +145,56 @@ def _run(app: App):
...
+                "Use the filter 'Not on This Device' and use CTRL+A to "
+                "make this easier.\n"
```
thw26, line 160: You need to add direction here to hit the "Download" button or whatever it is called in Logos.

ou_dedetai/repair.py
```diff
@@ -98,6 +145,56 @@ def _run(app: App):
...
+                "After downloading all your resources, wait until after it indexes "
```
thw26, line 161: change end of line to say "wait until after {flproduct} indexes your library "

ou_dedetai/repair.py
```diff
@@ -98,6 +145,56 @@ def _run(app: App):
...
+                "before using the application - some features may crash if they are "
```
thw26, line 162: Change the hyphen to an emdash: — and remove the surrounding whitespace.

ou_dedetai/repair.py
```diff
@@ -98,6 +145,56 @@ def _run(app: App):
...
+
+            if not app.approve(question=question, context=context):
+                return
+            app.status(f"Bypassing first-run dialog for {persistent_config.faithlife_product}") #noqa: E501
+            logos_appdata_dir = app.conf._logos_appdata_dir
+            logos_user_id = app.conf._logos_user_id
+
+            if logos_appdata_dir is None:
+                # This shouldn't happen - we use this dir when detecting this failure
+                app.status("Failed to recover first time resource download - can't find Logos dir") #noqa: E501
```
thw26, line 174: Change the hyphen to an emdash: — and remove the surrounding whitespace, or change it to a colon.

ou_dedetai/repair.py
```diff
@@ -98,6 +145,56 @@ def _run(app: App):
...
+                time.sleep(5)
+                return
+            if logos_user_id is None:
+                # This shouldn't happen - we use this dir when detecting this failure
+                app.status("Failed to recover first time resource download - can't find Logos user Data dir") #noqa: E501
```
thw26, line 179: See above.

You might want to add something like, "Did you log in" to the error message, or, "Did you delete a directory?" to these two errors; as your comment notes, we shouldn't need these checks, but good to have them just in case. As we should be able to find this dir so long as they've logged in once. This gives the user some direction rather than waiting for us to read the log and ask them what they did.

ou_dedetai/repair.py
```diff
@@ -98,6 +145,56 @@ def _run(app: App):
...
+                time.sleep(5)
+                return
+            local_user_prefrences_path = Path(logos_appdata_dir) / "Documents" / logos_user_id / "LocalUserPreferences" / "PreferencesManager.db" #noqa: E501
```
thw26, line 182: We may want an error as above in case the db is not found.

ou_dedetai/repair.py
```diff
@@ -98,6 +145,56 @@ def _run(app: App):
...
+            ou_dedetai.utils.execute_sql(
```
thw26, line 183: See comments on adding a db.py module.

We have a couple other bits of code in the works, sitting on the backburner, which also interact with the Logos sqlite DBs. It would be good to put these in their own module so that we don't have to retrieve them at a later time.

ou_dedetai/repair.py
```diff
@@ -98,6 +145,56 @@ def _run(app: App):
...
+                ]
+            )
+            # XXX: leave a note to future us to let us know the user skipped this during the installtion.
+            # Useful if this operation has side effects in the future, being able to tell that the user did this WAYYYY back when they first installed logos
+
+            app.status(
+                f"Recovery attempt of {app.conf.faithlife_product} complete. "
+                f"{app.conf.faithlife_product} should now launch directly, "
+                "bypassing first time resource download dialog."
+            )
```
thw26, line 196: We might want to include our instructions here telling the user to download their library.

thw26 (changes requested):
I or Nate need to test.

ou_dedetai/repair.py
ctrlaltf24: The sql call will throw an exception that will be caught by the surrounding try/catch. It's still handled - that's what the catch all was for

ctrlaltf24:
Reduced scope of PR drastically due to potential long term effects of skipping this dialog. Still valuable to merge due to the database.py changes

thw26:
Is the intent of this code in part to resolve things like #86 or just laying foundation now?

ctrlaltf24:
Nope it won't fix that issue (that issue just needs a app.is_installed), more so laying the ground work for more complex error state checking. Figured did all the work/code/testing to figure out when the first run dialog would show, it's worth having around

thw26 (approved):

Issue #197, opened 2024-10-14
Fix Backup and Restore


thw26:
With the release of Beta, a lot of the Backup and Restore functionality has been fixed so that the code works, but a few items need to be fixed before letting users run wild with it.

- [ ] Disable backup and restore if Logos or Indexing is running
- [ ] Provide the size of backup as a human-readable value to the user for confirmation
- [ ] Provide the size of target dir as a human-readable value to the user for confirmation
- [ ] Determine how to provide progress to the user, as the current comment-outed code is not triggering in the TUI

Issue #228, opened 2024-11-13
Switch TUI to urwid


thw26:
https://github.com/urwid/urwid

The current TUI is in progress of creating a high level curses/dialog API. Switching to urwid would let us drop tui_screen.py, tui_curses.py, and tui_dialog.py, and it would also let us drastically simplify tui_app.py and likely launching the TUI app from main.py.

This would in turn reduce our custom code maintenance but introduce another project as a dependency.

I believe this transition would be worth the effort, particularly surrounding our desires in #147 and in #206.

thw26:
Or see also https://click.palletsprojects.com/en/stable/

thw26:
This is particularly driven by a desire for a file browser in path selection. Here's another example:

https://gitlab.com/tripcodeq7/xplore_file_browser

thw26:
I found the Textual Python library. This is a great tool for TUI.

https://github.com/Textualize/textual

I have worked with this library on another project and it is excellent for TUI tools. Reopening this as I think this may be a decent option.

It works well with async code, unlike our current library.

PR #345, opened 2025-02-26
fix: backup cli

n8marti:
Get backup and restore working for CLI (prep for getting general backup/restore working)
- [x] Split backup/restore into own module and Include unittests
- [x] Ensure backup works to fixed and removable drives (snap pkg needs additional permission from user for removable drives)
- [x] Ensure restore works on clean Logos install
- [x] Ensure restore works on existing Logos install (i.e. overwriting existing data)
- [x] Re-publish CLI backup/restore options

### DIFF
--- a/ou_dedetai/app.py
+++ b/ou_dedetai/app.py
@@ -228,12 +228,40 @@ class App(abc.ABC):
                 if None that signifies we can't track the progress.
                 Feel free to implement a spinner
         """
+
+        def get_progressbar(pct, w=10, suffix=''):
+            suffix = f" {int(pct):>3}%"
+            # end = '\r'
+
+            ci = '\u23b8'
+            cf = '\u23b9'
+            d = '█'
+            u = ' '
+            d_ct = min(int(w*pct/100), w)
+            u_ct = min(int(w - d_ct - 1), w - 1)
+
+            bar = ''
+            if d_ct == 0:
+                bar += ci + u*(u_ct - 1) + cf
+            elif d_ct < w:
+                bar += str(d*d_ct) + str(u*u_ct) + cf
+            else:
+                bar += str(d*d_ct)
+            bar += suffix #+ end
+            return bar
+
         # De-dup
         if message != self._last_status:
+            bar = ''
+            if percent is not None:
+                bar = f"{get_progressbar(percent)}"
             if message.endswith("\r"):
-                print(f"{message}", end="\r")
+                print(f"{bar}{message}", end="\r")
             else:
                 print(f"{message}")
+                if bar:
+                    print(bar)
+        
 
     @property
     def superuser_command(self) -> str:
--- a/ou_dedetai/backup.py
+++ b/ou_dedetai/backup.py
@@ -1,25 +1,30 @@
+import abc
 import queue
 import logging
 import shutil
 import time
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 from typing import Tuple
 from ou_dedetai import constants
 from ou_dedetai import utils
 from ou_dedetai.app import App
 
 
-class BackupBase:
+class BackupBase(abc.ABC):
     DATA_DIRS = ['Data', 'Documents', 'Users']
 
-    def __init__(self, app: App) -> None:
+    def __init__(
+        self,
+        app: App,
+        mode: str,
+    ) -> None:
         self.app = app
-        self.mode = None
-        self.source_dir = None
-        self.data_size = None
-        self.destination_dir = None
-        self.destination_disk_used_init = None
+        self.mode = mode
+        self._destination_dir: Optional[Path] = None
+        self._source_dir: Optional[Path] = None
+        self.data_size = 1
+        self.destination_disk_used_init: Optional[int] = None
         self.q: queue.Queue[int] = queue.Queue()
         if not self.app.approve(f"Use existing backups folder \"{self.app.conf.backup_dir}\"?"): # noqa: E501
             # Reset backup dir.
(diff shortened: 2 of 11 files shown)


n8marti:
We might need/want to backup/restore Logos registry keys, as well as user data folders.

ou_dedetai/app.py
```diff
@@ -228,12 +228,40 @@ def _status(self, message: str, percent: Optional[int] = None):
...
+                bar += ci + u*(u_ct - 1) + cf
+            elif d_ct < w:
+                bar += str(d*d_ct) + str(u*u_ct) + cf
+            else:
+                bar += str(d*d_ct)
+            bar += suffix #+ end
+            return bar
+
         # De-dup
         if message != self._last_status:
```
ctrlaltf24, line 254: Should the bar print even if the status didn't change?

We may consider changing the Downloading.... status to use a percentage instead of adding dots

ou_dedetai/control.py
```diff
@@ -4,152 +4,152 @@
...
 import shutil
-import time
+# import time
 from pathlib import Path
 
 from ou_dedetai.app import App
 
 from . import system
-from . import utils
+# from . import utils
```
ctrlaltf24, line 16: ?

ou_dedetai/backup.py
```diff
@@ -0,0 +1,138 @@
...
+    q: queue.Queue[int] = queue.Queue()
+    message = "Calculating backup size…"
+    app.status(message)
+    i = 0
+    t = app.start_thread(utils.get_folder_group_size, src_dirs, q)
+    try:
+        while t.is_alive():
+            i += 1
+            i = i % 20
+            app.status(f"{message}{"." * i}\r")
```
ctrlaltf24, line 82: nit: use progress in status rather than adding dots

ou_dedetai/backup.py
```diff
@@ -0,0 +1,138 @@
...
+    t.join()
+    src_size = q.get()
+    if src_size == 0:
+        app.exit(f"Nothing to {mode}!")
+
+    # Set destination folder.
+    if mode == 'restore':
+        if not app.conf.logos_exe:
+            app.exit("Cannot restore, Logos is not installed")
+        dst_dir = Path(app.conf.logos_exe).parent
```
ctrlaltf24, line 96: Should this use logos app data Dir instead?

ctrlaltf24 (commented):
Nothing new, some thoughts feel free to ignore, I know you're still working

ou_dedetai/backup.py
n8marti: I had the same question, but I think the Logos *app* should be installed before trying to restore a backup. Otherwise installing Logos over existing data directories might clobber them? Not sure. Also, is `app.conf.logos_exe` generic enough to be defined as the Verbum exe path as well?

n8marti:
Consider a file chooser for setting Backup dir:
The TUI one is a text field with mkdir. [tui_app.py:L774](https://github.com/FaithLife-Community/OuDedetai/blob/68d21520b6c49c06abe4871117688eeca623876b/ou_dedetai/tui_app.py#L774)

thw26:
> Consider a file chooser for setting Backup dir: The TUI one is a text field with mkdir. [tui_app.py:L774](https://github.com/FaithLife-Community/OuDedetai/blob/68d21520b6c49c06abe4871117688eeca623876b/ou_dedetai/tui_app.py#L774)

cf. #228 for links to some potential curses file browsers.

ctrlaltf24:
Ran additional tests on 78aeef0cc450ff9e2e9b86bd3b6f7e6b9e73c4f6:
- Backup before user login -> success
- Backup before user login, then login, then restore backup before login -> not logged in
- Backup with logos account one, restore backup before login, login with second logos account, restore logos account one backup -> logos account one restored
- Backup one one directory, change install dir, install into that directory, restore -> success

ctrlaltf24 (approved):
Approved, although would like to see https://github.com/FaithLife-Community/OuDedetai/pull/360 merged into this first.

ou_dedetai/backup.py
ctrlaltf24: > Also, is app.conf.logos_exe generic enough to be defined as the Verbum exe path as well?

Yup

ou_dedetai/utils.py
```diff
@@ -276,31 +276,37 @@ def find_installed_product(faithlife_product: str, wine_prefix: str) -> Optional
...
 
-def enough_disk_space(dest_dir, bytes_required):
+def enough_disk_space(dest_dir, bytes_required) -> bool:
     free_bytes = shutil.disk_usage(dest_dir).free
     logging.debug(f"{free_bytes=}; {bytes_required=}")
     return free_bytes > bytes_required
 
 
-def get_path_size(file_path):
+def get_path_size(file_path: Path|str) -> int:
```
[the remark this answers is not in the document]
n8marti: sure. Is it significantly faster? Less memory and/or CPU intensive? Or is there some other benefit to `os.walk` over `Path.rglob`?

ou_dedetai/tui_app.py
```diff
@@ -861,9 +862,9 @@ def do_backup(self):
         self.todo_e.wait()
         self.todo_e.clear()
         if self.tmp == "backup":
-            control.backup(self)
+            backup.backup(self)
         else:
-            control.restore(self)
+            backup.restore(self)
```
thw26, line 867: We need to re-enable backup and restore in the menu below, lines 983–984.

thw26 (changes requested):

ou_dedetai/tui_app.py
thw26: Perhaps I misunderstood, but this PR is not intended to fix #197, only partially so, correct?

ctrlaltf24: The way I understood it was an improvement, related to #197, but not aimed at fixing it. I looked to the PR description to determine scope rather than the issue

ou_dedetai/utils.py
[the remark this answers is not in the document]
ctrlaltf24: doesn't require recursion is the main benefit which may be inconsequentially faster

PR #349, opened 2025-03-01
feat: TUI: add a footer window; split console into header and console windows

### LINKED ISSUE #342: TUI: Add third window as footer
Currently the TUI is composed of two windows: the header or console window and the main window. Add a third window as a footer, below the main window, in which would be things like a progress bar and the current position on a menu.

Previously I had wanted to implement a progress bar in the TUI but the question had been how. I hadn't considered adding a third window. For that matter, the window could also be split up so that there is a header and a console window.

One benefit of this is if I can get things moving ahead with Arm, on smaller screens I will want things to be more adaptive, and having more windows would let me hide a window or otherwise on a smaller screen.

thw26:
Fixes #342.

Besides declaring new variables for each of the new windows, this also renames the console window to header and the menu window to main. As part of this, some of the functions needed to have the appropriate screen passed to them rather than calling that screen from the App. I tried a branch where I removed the App dependency from tui_screen.py and tui_curses.py but I was unable to get the drawing correct. I believe I need to rework that branch as I think there is some bad tracking with the new variables.

I modularize some of the bigger functions and simplify some of the functions where I could implement logic. I also swap some of the static numbers for their appropriate variable.

I also added the ability to navigate TUI menus with left and right arrow keys. What sparked the add was that when the TUI is small enough to only display one option in the menu, up and down no longer make sense. Now the user can use either set of keys.

There is more refactoring that can be done, such as passing the footer contents to the footer screen rather than declaring them within the footer. Those were things I had tackled in the refactor away from an app dependency. I'll look at adding them in again in the future in smaller PRs.

### DIFF
--- a/ou_dedetai/tui_app.py
+++ b/ou_dedetai/tui_app.py
@@ -250,7 +250,7 @@ class TUI(App):
         self.header_window_ratio = 0.10
         self.console_window_ratio = 0.15
         self.footer_window_ratio = 0.10
-        self.main_window_ratio = 0.55  # Intentionally short his by 10% to avoid hidden lines
+        self.main_window_ratio = 0.55  # Intentionally short this by 10% to avoid hidden lines
 
         self.set_header_window_dimensions()
         self.set_console_window_dimensions()
--- a/ou_dedetai/tui_app.py
+++ b/ou_dedetai/tui_app.py
@@ -75,7 +75,7 @@ class TUI(App):
 
         # Default height and width to something reasonable so these values are always
         # ints, on each loop these values will be updated to their real values
-        self.window_height_min = 20
+        self.window_height_min = 11
         self.window_height = self.window_width = 80
         self.header_window_height = self.header_window_width = 80
         self.console_window_height = self.console_window_width = 80
@@ -110,7 +110,7 @@ class TUI(App):
         # corresponding @property functions
         self._main_screen: Optional[tui_screen.MenuScreen] = None
         self._active_screen: Optional[tui_screen.Screen] = None
-        self._header: Optional[tui_screen.ConsoleScreen] = None
+        self._header: Optional[tui_screen.HeaderScreen] = None
         self._console: Optional[tui_screen.ConsoleScreen] = None
         self._footer: Optional[tui_screen.FooterScreen] = None
         # End internal property values
@@ -263,7 +263,7 @@ class TUI(App):
         header_window_start = 0
         console_window_start = self.header_window_height
         main_window_start = self.header_window_height + self.console_window_height + 1
-        footer_window_start = self.header_window_height + self.console_window_height + self.main_window_height + 2
+        footer_window_start = self.window_height - self.footer_window_height - 1
 
         self.header_window = curses.newwin(self.header_window_height, curses.COLS, header_window_start, 0)
         self.console_window = curses.newwin(self.console_window_height, curses.COLS, console_window_start, 0)
--- a/ou_dedetai/tui_app.py
+++ b/ou_dedetai/tui_app.py
@@ -75,30 +75,38 @@ class TUI(App):
 
         # Default height and width to something reasonable so these values are always
         # ints, on each loop these values will be updated to their real values
+        self.window_height_min = 21
         self.window_height = self.window_width = 80
-        self.main_window_height = self.menu_window_height = 80
+        self.header_window_height = self.header_window_width = 80
+        self.main_window_height = self.main_window_width = 80
+        self.footer_window_height = self.footer_window_width = 80
         # Default to a value to allow for int type
-        self.main_window_min: int = 0
-        self.menu_window_min: int = 0
+        self.header_window_height_min: int = 0
+        self.main_window_height_min: int = 0
+        self.footer_window_height_min: int = 0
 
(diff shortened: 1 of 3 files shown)


ou_dedetai/tui_screen.py
```diff
@@ -79,19 +77,64 @@ def display(self):
...
+        return "Curses Footer Screen"
+
+    def display(self):
+        if self.stdscr is None:
+            raise Exception("stdscr should be set at this point in the footer screen."
+                            "Please report this incident to the developers")
+        self.stdscr.erase()
+
+        footer_text = "By the FaithLife Community"
+        footer_hr = "---"
```
ctrlaltf24, line 132: unused
```suggestion
```

ou_dedetai/tui_app.py
```diff
@@ -81,30 +81,44 @@ def __init__(self, stdscr: curses.window, ephemeral_config: EphemeralConfigurati
...
         self.current_page: int = 0
         self.total_pages: int = 0
+        self.menu_bottom: int = 0
 
         # Start internal property variables, shouldn't be accessed directly, see their 
         # corresponding @property functions
-        self._menu_screen: Optional[tui_screen.MenuScreen] = None
+        self._main_screen: Optional[tui_screen.MenuScreen] = None
         self._active_screen: Optional[tui_screen.Screen] = None
+        self._header: Optional[tui_screen.ConsoleScreen] = None
```
ctrlaltf24, line 119:
```suggestion
        self._header: Optional[tui_screen.HeaderScreen] = None
```

ou_dedetai/tui_app.py
```diff
@@ -162,63 +176,117 @@ def active_screen(self, value: tui_screen.Screen):
...
 
-    def set_window_dimensions(self):
-        self.update_tty_dimensions()
-        curses.resizeterm(self.window_height, self.window_width)
-        self.main_window_ratio = 0.25
+    def set_header_window_dimensions(self):
+        self.header_window_height_min = 2
+        self.header_window_height = min(max(
+            int(self.window_height * self.header_window_ratio), self.header_window_height_min
+        ), 4)
```
ctrlaltf24, line 224: nit: why is this dynamic? When it's set to static when resizing space doesn't appear between the welcome and the console. Feels odd to have this dynamic
```suggestion
        self.header_window_height = 2
```

ou_dedetai/tui_screen.py
```diff
@@ -79,19 +77,64 @@ def display(self):
...
+    def __str__(self):
+        return "Curses Footer Screen"
+
+    def display(self):
+        if self.stdscr is None:
+            raise Exception("stdscr should be set at this point in the footer screen."
+                            "Please report this incident to the developers")
+        self.stdscr.erase()
+
+        footer_text = "By the FaithLife Community"
```
ctrlaltf24, line 131: nit: Wouldn't this make more sense either in the same line as page (but justified right) or next to Welcome to Ou Dedetai? Feels strange as the last line

thw26: With the way curses draws, I don't think it'd be the funnest to code this suggestion. I debated it myself but we would need to implement a few things as part of the draw, or even a fifth window that splits the footer into a left and right column. It is easier at the moment to add a line.

Using a suggested refactor, I was able to get the screen minimum height down to 11 again. If in the future we program the draw to be able to do your suggestion, we could get it down to 10.

ou_dedetai/tui_app.py
thw26: This is dynamic because the title lines can wrap, causing the title to be as much as 4 lines.

thw26: This brings up a good question though: when the screen is quite narrow, the title lines wrap. If the screen is small enough, the second title line is hidden because it needs at least 3 lines to display the first part of the subtitle. But in a small enough screen, the header window is only 2 lines. What approach should we take to handle this?

thw26: In a small enough window (width-wise), the height of the terminal would need to be a minimum of 40 to display the contents of a wrapped TUI subtitle.

thw26: We could set this to a constant value of 4, but it makes for some funny white space at certain sizes.

ctrlaltf24: Are you talking about when the window is this small?
```
  Welcome to Ou Dedetai
  4.0.0-beta.9 (stable)
  Logos Version: 33.1.0.6
  (stable)
```

thw26: Correct, or thereabout

ctrlaltf24: Do we really care about all that information if the screens that small?

thw26: Of course not, but I wasn't intending to make this into a responsive refactor with properly hiding things.

ctrlaltf24: Could we hide the last two lines on small screens? To keep it 2 lines always? Simplify things?

thw26: Hmm... My preferred way of doing that requires a major refactor. But there is an easy way. Let me test something for it.

ctrlaltf24 (approved):

Issue #274, opened 2025-01-23
Add a OD Reporting Button


thw26:
Similar to #103, we should have a basic option that outputs a user's system and config, sort of like our own version of `inxi`. Then we could quickly gather data on a user's system. For ease of getting support

n8marti:
To make the UI flow more explicit, are you proposing a button/menu option that would do something like the following?:

- Gather relevant system and installation info
- Export it somehow (to a text file?)
- Save it somewhere (to user's `$HOME`?)
- Maybe also provide the user with relevant details about where to send the file?
- Maybe automatically send the report somewhere? (new email address maybe?)

Alternatively, we could maybe generate this report on startup and always save it to `~/.local/state/FaithLife-Community`, and just overwrite it each time the app runs? Then the user just needs to find it and send it in case of a problem?

Issue #103, opened 2024-05-20
Create an upload log option


thw26:
For easier access and more GUI controls, we could create an option to upload the load automatically to some location. This is aimed at new Linux users getting the support they need.

See #153 for prep work for this.

Files to include:
- OD config
- OD log
- wine.log
- glxinfo
- free
- handful of envs (like which desktop environment the user is using)
- /etc/os-release
- wine --version

ctrlaltf24:
Also increase viability of support options:
- matrix/telegram
- github issues

PR #353, opened 2025-03-01
feat: generate support bundle

ctrlaltf24:
Fixes: #274, #103

Still need to test one last time with all 6 combinations (GUI, TUI, CLI), (default and customer dirs), as well as all 3 buttons

### DIFF
--- a/ou_dedetai/app.py
+++ b/ou_dedetai/app.py
@@ -83,7 +83,11 @@ class App(abc.ABC):
                 return answer
             if PROMPT_OPTION_DIRECTORY in options and Path(answer).is_dir():
                 return answer
-            if PROMPT_OPTION_NEW_FILE in options and Path(answer).parent.is_dir():
+            if (
+                PROMPT_OPTION_NEW_FILE in options
+                and not Path(answer).is_dir()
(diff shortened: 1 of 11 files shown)


thw26 (changes requested):
File generation works.

In TUI, code currently errors if you select "Save As" and don't enter a dir. Reports it can't perform the action because the target is a directory. Needs a default dir specified.

If I select Return to Main Menu on the choose support save location page (first page) and on the support link page, I also get an error on attempting to return rather than choosing a support link.

From the first, the error appears to be non-breaking. Shows up as a "raise ReturningToMainMenu" error, line 813 in _ask in tui_app.py

From the second, it breaks the menu, giving the same error, but the menu becomes broken and requires a Ctrl+C.

thw26:
If you would, please move the menu item for support to after "Change Color Scheme". It might be nice to have the option of not generating the support archive but bypassing it to get to the links.

ctrlaltf24:
> It might be nice to have the option of not generating the support archive but bypassing it to get to the links.

In order to get support they'll need the bundle. However good suggestion of some way of seeing the links, added another option to just see them

thw26 (approved):

PR #360, opened 2025-03-06
refactor: use abstract base class in backup

ctrlaltf24:
rather than initializing destination_dir to none, then setting it, use abstract functions to populate them.

Also standardized on logos_app_data_dir rather than using logos_app_data_dir in one operation, then logos_exe.parent in another

Resolves type issues reported by mypy

Tested:
- installed Logos, logged in, downloaded resources, made backup, deleted Data/Documents/Users, restored -> success

### DIFF
--- a/ou_dedetai/backup.py
+++ b/ou_dedetai/backup.py
@@ -1,25 +1,30 @@
+import abc
 import queue
 import logging
 import shutil
 import time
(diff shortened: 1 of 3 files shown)


n8marti (approved):
This looks fine to me.

PR #362, opened 2025-03-07
feat: add uninstall button

ctrlaltf24:
Fixes: https://github.com/FaithLife-Community/OuDedetai/issues/233

Removes the --remove-install-dir cli argument and adds --uninstall

Tested:
- CLI (Yes, No), (No), (Yes, Yes)
- GUI (No), (Yes, No, Then install again) (Yes, Yes). Note: This will re-create a clean config and cache after uninstall due to GUI enforcing settings
- TUI (Return to main menu), (Yes, Return to Main Menu), ((Yes, No), Install, (Yes, Yes))

### DIFF
--- a/ou_dedetai/tui_app.py
+++ b/ou_dedetai/tui_app.py
@@ -522,12 +522,14 @@ class TUI(App):
         self.choice_q.put("Return to Main Menu")
 
     def main_menu_select(self, choice):
+        original_assume_yes = self.conf._overrides.assume_yes
         def _install():
             try:
                 installer.install(app=self)
             except UserExitedFromAsk:
                 pass
             finally:
+                self.conf._overrides.assume_yes = original_assume_yes
                 self.go_to_main_menu()
 
         if choice is None or choice == "Exit":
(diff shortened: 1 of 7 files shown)


ou_dedetai/tui_app.py
```diff
@@ -633,63 +635,72 @@ def wineconfig_menu_select(self, choice):
...
+                )
+            elif choice == "Install ICU":
+                self.reset_screen()
+                wine.enforce_icu_data_files(self)
+                self.go_to_main_menu()
+            elif choice.endswith("Logging"):
+                self.reset_screen()
+                self.logos.switch_logging()
+                self.go_to_main_menu()
+            elif choice == "Uninstall":
```
thw26, line 698: First Issue: I have an install located in .local/share/FaithLife-Community/oudedetai, but because either 1) there is no config for it, or 2) it is a partially completed install (i.e., app.is_installed() reports false) it is not recognized as existing. This option does not show up therefore, even though the default directory has contents in it. Is there a way for us to detect this kind of situation properly? And if so, what is the proper thing to do with our UI?

thw26, line 698: Second Issue: I added a config file with the proper directory set as the install dir, but because the install is only partially completed, this option does not show. Is there a way to detect this kind of situation properly? And if so, what is the proper thing to do with our UI?

Here is the config I used:

```
{
    "app_release_channel": "stable",
    "curses_color_scheme": "System",
    "faithlife_product": "Logos",
    "faithlife_product_release_channel": "stable",
    "faithlife_product_version": "10",
    "install_dir": "/home/thwright/.local/share/FaithLife-Community/oudedetai",
    "wine_binary": "data/bin/wine-staging_10.0-rc5-x86_64.AppImage"
}
```

ou_dedetai/tui_app.py
```diff
@@ -969,6 +980,7 @@ def set_utilities_menu_options(self):
                 "Remove Library Catalog",
                 "Remove All Index Files",
                 "Install ICU",
+                "Uninstall"
```
thw26, line 983: I think this should be moved to line 995, after restore.

thw26 (changes requested):

ou_dedetai/tui_app.py
ctrlaltf24: Hmm good question, should we always show this if any of the dots we would remove have contents?

ou_dedetai/tui_app.py
ctrlaltf24: My rational for having it that high was so the user could see it (but not so high they accidentally hit it) . Moving it all the way down there doesn't show on my 1/4 screen TUI - 17 lines. If the user does want to uninstall - I'm not sure they'd read all these options. I'm open to moving it, but would prefer a place the user could see

PR #364, opened 2025-03-07
fix: actually move progress bar as we download

ctrlaltf24:
This assumes that there is a 1:1 relationship between steps and download jobs, which is presently true. If at some point in the future it is no longer true the worst that'll happen is the progress bar will appear to go backwards.

Tested:
- GUI clean install w/ clearing cache, observe progress bar moving as resource is downloading with human readable name
- CLI clean install w/ clearing cache, observed characters moving (there is a double line as the file is downloading (this may be fixed in the latest commit, couldn't test due to GH rate limit))

### DIFF
--- a/ou_dedetai/network.py
+++ b/ou_dedetai/network.py
@@ -379,7 +379,7 @@ def _net_get(url: str, target: Optional[Path]=None, app: Optional[App] = None):
     logging.debug(f"Download destination: {target}")
     target_props = FileProps(target)  # sets path and size attribs
     if app and target_props.path:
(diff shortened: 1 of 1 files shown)


thw26:
Perfect! Now I want it in the TUI!

thw26 (approved):

thw26:
Your most recent commit message says WIP. I am waiting for a commit message amend before I merge.

ctrlaltf24:
Wanted to retest CLI clean install w/ clearing cache to see if there is a double line as the file is downloading

But it's safe to squash merge as-is

PR #370, opened 2025-03-09
feat: add progress bar to CLI

ctrlaltf24:
It should be noted this restricts the CLI to only working on *NIX system - as it uses curses

Tested:
- Ran clean install, saw progress bar and no artifacts, and after OD completed, could interact with terminal again

### DIFF
--- a/ou_dedetai/app.py
+++ b/ou_dedetai/app.py
@@ -208,8 +208,7 @@ class App(abc.ABC):
         """A status update
         
         Args:
-            message: str - if it ends with a \r that signifies that this message is
-                intended to be overrighten next time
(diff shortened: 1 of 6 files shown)


thw26:
This needs the fixes found in #379. Please rebase on latest main for testing.

ctrlaltf24:
> Any thoughts on avoiding curses?
> 
> 
> 
> https://stackoverflow.com/questions/3173320/text-progress-bar-in-terminal-with-block-characters

That's only a single line. If we want a global progress bar for the installation we need to go up a line - i.e. curses

thw26 (approved):

PR #372, opened 2025-03-09
test: expand integration test

ctrlaltf24:
Now the integration test:
- Tests clean install
- Tests first time resource download
- Tests the following tools/guides to ensure they don't crash
  - open bible
  - copy bible verses
  - factbook
- ensures crash detection logic works (with a test!)

Lots of room to expand this, and obvious integrate into unittest somehow (we'll need to be able to filter these tests separately as their design goals is different). These tests are designed for: Does OD install Logos in such a way that it doesn't crash when doing normal tasks? While unittest is designed for: does the python do what we expect it to do?

This file should also be broken up when it's integrated, kept as one file for now, but it will be split.

### DIFF
--- a/tests/integration.py
+++ b/tests/integration.py
@@ -93,7 +93,7 @@ class OuDedetai:
     def isolate_files(self):
         if self._temp_dir is not None:
             shutil.rmtree(self._temp_dir)
-        # XXX: this isn't properly cleaned up. Context manager?
+        # FIXME: this isn't properly cleaned up when tests fail. Context manager?
         self._temp_dir = tempfile.mkdtemp()
         self.config = Path(self._temp_dir) / "config.json"
         self.install_dir = Path(self._temp_dir) / "install_dir"
@@ -171,7 +171,7 @@ class OuDedetai:
             self.stop_app()
         except Exception:
             pass
-        # XXX: Ideally the uninstall operation would automatically stop the app.
+        # FIXME: Ideally the uninstall operation would automatically stop the app.
         # Open an issue for this.
         self.run(["--uninstall", "-y"])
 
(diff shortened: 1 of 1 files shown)


ctrlaltf24:
Going to switch gears now and work on a follow-on PR to integrate this into CI. 

This can merge-as is (so you guys can use it to test other PRs in the meantime)

tests/integration.py
```diff
@@ -130,98 +165,889 @@ def run(self, *args, **kwargs):
...
+
+    def __init__(
+        self,
+        pre_input_tasks: Callable[[], None]
+    ):
+        self.pre_input_tasks = pre_input_tasks
+    
+    @classmethod
+    def detect(cls, pre_input_tasks: Callable[[], None]) -> "DisplayServer":
+        """Detects the current running Display server and returns and interface
```
n8marti, line 524: typo: "...and returns *an* interface"

tests/integration.py
```diff
@@ -130,98 +165,889 @@ def run(self, *args, **kwargs):
...
+              "wouldn't hit the issue in question.")
+        return
+    if FaithLifeProduct.name() == "Verbum":
+        print("Skipping test for winedbg detection in test code since it doesn't "
+              "occur in Verbum.")
+        return
+
+    # Sabotage! For the sake of a crash. Still needs testing to force a crash. 
+    # We may want to have a negative test for this to ensure our logic to detect
+    # logos idn't crash still functions later on in time.
```
n8marti, line 955: typo: "logos *didn't* crash..."

n8marti (changes requested):
This is awesome! Good, thorough, start on testing most basic functionality. Only a couple of requests to fix typos.

Do you think it would be possible to implement this in a GH action (down the road, not necessarily now)? As in, use something like ubuntu-latest, then install a desktop environment, then run the tests?

ctrlaltf24:
Yup that's the goal! That's considerably harder than expanding this, so made it into two separate PRs. I'll also need to add backup/restore tests, as well screenshot functionality, as well as triggers for the full integration test dream

PR #382, opened 2025-03-11
fix: incorrect TUI ratio value

thw26:
Following #379, which I merged without fully testing in order to be able to test other PRs, I found there was a regression in the PR. This turned out to be an incorrectly assigned value.


n8marti (approved):
Looks pretty straightforward to me.

PR #383, opened 2025-03-18
chore(release): update version to 4.0.0-beta.11

ctrlaltf24:
Tested:
- Clean install, running logos shows login screen for Logos 40.2
- launch TUI, shown main menu
- launch GUI, shown control panel

Post-release:
- Make sure GLIBC version is old enough
- Make sure oudedetai x86 binary is at the top and named oudedetai (may need to download the binary, rename it locally, then re-upload). This may or may not matter, our code MAY assume one name over the other.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+- 4.0.0-beta.11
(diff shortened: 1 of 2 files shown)


n8marti (approved):
Looks good.

PR #384, opened 2025-03-19
feat: support Stop app operation on GUI

ctrlaltf24:
The TUI already has this behavior

Tested (40.2 clean install pre-login):
- Started OD with app installed, but not running -> Run Logos shown
- When "Run Logos" is hit, button switches to "Stop Logos"
- When Logos was launched using OD, but is closed independently of OD, button switches to "Run Logos" after the app closes
- "Run App" -> Switches to "Stop App", when "Stop App" is clicked, it stops Logos and switches to "Run Logos", when "Run Logos" is hit again, it launches
- GUI when Logos isn't installed no errors seen on console
- When Logos is installed via the GUI, button switches from "Install" to "Run Logos". Hitting that switches to "Stop Logos", hitting that switches to "Run Logos" which launches Logos

### DIFF
--- a/ou_dedetai/gui_app.py
+++ b/ou_dedetai/gui_app.py
@@ -3,6 +3,7 @@
 #   - https://tkdocs.com/
 #   - https://github.com/thw26/LogosLinuxInstaller/blob/master/LogosLinuxInstaller.sh
 
+import copy
 import logging
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


n8marti (approved):
Looks good. Merge it in if you feel like you've done sufficient testing.

PR #399, opened 2025-04-12
Replace p7zip with 7zip on Arch (Fixes #398)

### LINKED ISSUE #398: 7-Zip installation causes issues on Arch Linux
Seems as if the p7zip package was moved to 7zip on Arch Linux, and that the code has been changed for pamac but not for Arch or SteamOS. I'm trying to setup up a development environment on my system to fix this, but if someone can beat me to it, that would be appreciated.

DavidDGTNT:
I only fixed the Arch side of things as I don't have a way to test SteamOS, however, I would assume that it needs changed as well. This fixes issue #398, which I filed just before fixing it when I was unsure on whether or not I could get a dev environment set up.

### DIFF
--- a/ou_dedetai/system.py
+++ b/ou_dedetai/system.py
@@ -526,7 +526,7 @@ def get_package_manager() -> PackageManager | None:
             packages = (
                 "fuse2 "  # appimages
                 "binutils libwbclient samba wget "  # wine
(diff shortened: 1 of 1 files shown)


thw26 (approved):

thw26:
Thank you!

Issue #264, opened 2025-01-20
Add support for libronixdls (and logos4) urls


DavidDGTNT:
Certain websites (including my own) use the [Logos Reftagger](https://www.logos.com/reftagger) plugin, which has the capability of adding a "Open in Logos Bible Software" icon next to the reference. This attempts to open Logos Bible Software using the libronixdlx: format, which does not work with Ou Dedetai.

Expected behavior:
* Browse to a website with the Reftagger plugin installed and to a page with a Scripture reference. (e.g. the footer of [my blog](https://house.daviddgtnt.xyz))
* Click the ">L" button (with tooltip "Open in Logos Bible Software (if available)")
* Logos Bible Software opens

thw26:
Thank you for this! I use RefTagger on my sites.

There is probably a way for us to associate the links with Logos in a DE. We need to research those mechanisms, but our installer could make sure they are put in place.

thw26:
A general idea for how to make this work. Add the following at `~/.local/share/applications/libronixdls.desktop`:

```
[Desktop Entry]
Name=Libronix Keylink Handler
Exec=ADD LOGOS EXE CMD HERE "%u"
Type=Application
MimeType=x-scheme-handler/libronixdls;
NoDisplay=true
```

At CLI:, first add, then verify:

xdg-mime default libronixdls.desktop x-scheme-handler/libronixdls
xdg-mime query default x-scheme-handler/libronixdls

ctrlaltf24:
Criticality of this has now increased due to the login with Logos button using a logo4 formed url scheme. The same can be done for libronixdls. Testing for both logos4 and libronix...
```bash
echo > ~/.local/share/applications/logos4.desktop << EOF 
[Desktop Entry]
Type=Application
Name=Logos Scheme Handler
Exec=/home/user/.local/share/FaithLife-Community/oudedetai/oudedetai --wine "C:\\Users\\user\\AppData\\Local\\Logos\\Logos.exe" '%u'
StartupNotify=false
MimeType=x-scheme-handler/logos4;
EOF

update-desktop-database
xdg-mime default logos4.desktop x-scheme-handler/logos4
# Confirm registration if desired:
# cat ~/.config/mimeapps.list 
```

useful articles on the topic: 
- http://kb.mozillazine.org/Register_protocol
- https://learn.microsoft.com/en-us/windows/apps/develop/launch/handle-uri-activation

Issue #400, opened 2025-04-12
Consider UX of new 40.1 login flow


ctrlaltf24:
Now logins work like this:
0. Logos application starts for the first time showing a button saying log in
1.	Linux web browser opens
2.	Login screen given (if not already logged in)
3.	Page saying:
```
Redirecting you to Logos Bible Study. You are now signed in and can close this page.

If you aren’t redirected automatically, click here to try again.
```
4.	Click “click here" to try again
5.	Then click on the new option that appears:
```
Still having trouble? Click here and switch to Logos Bible Study, or contact support.
```
6.	After a couple seconds the Logos application window running in wine switches from its button to open the web browser to the first run dialog that either sets up the library or prompts you to choose how many resources are downloaded

Ideally the user would be informed at very least that step 5 is necessary. More ideal still would be make the extra unnecessary

86ul:
As https://github.com/FaithLife-Community/OuDedetai/issues/421 is referred to this, I would like to update my own bug report. After downloading, and indexing succesfully, Logos 43.0 can no longer connect to the server said it in a error message above/on the sync icon(?). Then I closed it, in order to restart the program, but it logged me out/login no longer worked. So that I tried to login once more via the browser, but now I got this error message: Could not read file logos4:AuthPortalCallback_284_dHsS07xujUyllTzq34kjMQ?oauth_token=[token1]&oauth_verifier=[token2]&utm_auth_action=signin. Token1 and Token2 are replaceing the real tokens for possible security issues related to my own FaithLife account...

EDIT: Several reboots later (for other reasons), I am able to login again. After logging in/authorizing (again) my e-mail account, I had to do the following:

Click on "OK" in the pop up error message
Click on "click here" in second sentence of the first "Nearly there!" message:
If you aren’t redirected automatically, click here to try again.

Again on "OK" in the (same?) pop up error message
Click on "Click here" in the third sentence of the second "Nearly there!" message:
Still having trouble? Click here and switch to Logos Bible Study, or contact support.

ctrlaltf24:
Fixed in v4.0.0-beta.12

PR #401, opened 2025-04-13
feat: support logos4 and libronixdls URL schemes

ctrlaltf24:
Open questions:
- snap - probably just needs a new .desktop file
- flatpak - maybe it's own desktop file

Tested (debian 12 only):
- built binary
- copied binary into clean VM
- installed Logos 40.1
- Launched Logos
- Hit "sign in" on Logos wine window
- Browser opened a window where I was asked wo "Authorize" Logos Bible Study (I was already signed in)
- Accepted the firefox prompt: "Allow this site to open the logos4 link with Logos URL Handler"?
- After saying yes, Logos said "Getting ready" and showed first run dialog as it did in 40.0 and below after logging in.
- After resources were downloaded went to a website in firefox that has a libronixdls URL (https://house.daviddgtnt.xyz/) and clicked the link
- Accepted the firefox prompt: "Allow this site to open the libronixdls link with Logos URL Handler"?
- Logos opened Ps 150:6 (which the link showed)!

Testing needed:
- SteamOS
- Manjaro (tested)
- Alpine (should delay until after #422)
- Fedora (already verified package name online)
- Archlinux  (should be the same as Manjaro, besides the package list)

Fixes: #264 #400

### DIFF
--- a/io.github.FaithlifeCommunity.OuDedetai.yml
+++ b/io.github.FaithlifeCommunity.OuDedetai.yml
@@ -99,6 +99,7 @@ modules:
       - type: file
         path: ./snap/gui/verbum.png
         dest-filename: verbum.png
+      # FIXME: Also support logos4 and libronixdls URL Schemes
       - type: inline
         dest-filename: oudedetai.desktop
         contents: |
--- a/ou_dedetai/config.py
+++ b/ou_dedetai/config.py
@@ -502,6 +502,9 @@ def get_logos_user_id(
     logos_appdata_dir: str
 ) -> Optional[str]:
     logos_data_path = Path(logos_appdata_dir) / "Data"
+    # If somehow the directory does not exist - like if the user removed it manually while trying to reset their data.
+    if not logos_data_path.exists():
+        return None
     contents = os.listdir(logos_data_path)
     children = [logos_data_path / child for child in contents]
     file_children = [child for child in children if child.is_dir()]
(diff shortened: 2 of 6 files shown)


n8marti:
It looks like snap support for this could be tricky (i.e. maybe not possible): https://forum.snapcraft.io/t/allow-snaps-to-register-new-mime-types/6467
But I don't yet fully understand all the steps. Are we solving 1 problem or 2 here? Specifically, are these the same or two separate things:
1. Logos confirming user login (presumably redirecting to a `.logos4` URL in the browser, which then opens Logos if that URL scheme has been registered at the system level)
1. Logos being opened when the user clicks on a `.libronixdls` URL in the browser.

It seems like they're the same programming problem, but just with two different MIME Types, right? In both cases they rely on the Ou Dedetai installer to register the relevant MIME Type at the system level so that when the URL is activated in the browser it knows to redirect to the app, right?

n8marti:
After reading some more about this and re-examining this PR, I'm now thinking this *will* be doable in the snap package. I don't think I see anything that involves registering a new MIME type at a system level, only doing so at the user level. And it looks like the command `xdg-mime` just adds a line to the relevant `mimetypes.list` file at the user level, which I think could be done no problem in the confined snap environment.

thw26:
Test on Manjaro. In general this works, but I had to click the extra buttons to return to Logos after the install and was not automatically redirected. Otherwise everything went well.

thw26 (approved):

PR #422, opened 2025-07-11
feat(wip): use wine from extracted appimage

### LINKED ISSUE #316: Consider packaging wine as a tarball rather than an appimage
Mounting an appimage requires fuse dependencies and some configuration on alpine. I'd be ideal if we didn't have to do this. If we package as a tarball (and the binaries are linked properly), we can use them instead, possibly with some LD_LIBRARY_PATH modifications. 

Preliminarily work here: https://github.com/ctrlaltf24/LogosLinuxInstaller/commit/d1469046400b37e24b6b7304789c04fdf9535aea

ctrlaltf24:
removes wine code logic

switches default to using the tarball of the appimage instead to avoid the fuse dependency
- this doesn't yet work inside of distrobox but does work on my debian host. Not pushing yet as there is clearly something I'm not understanding as to how the appimage gets mounted. I'd prefer if this was a pure binary/LD_Library path manipulation rather than having to use the wrapper from the appimage

begins to start writing ci around distrobox

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ venv/
 *.egg-info
 .vscode
 /.flatpak-builder
+/distrobox
--- a/ou_dedetai/cli.py
+++ b/ou_dedetai/cli.py
@@ -107,7 +107,7 @@ class CLI(App):
         self.logos.switch_logging()
 
     def update_latest_appimage(self):
-        utils.update_to_latest_recommended_appimage(self)
+        installer.update_to_latest_wine(self)
 
     def update_self(self):
         utils.update_to_latest_lli_release(self)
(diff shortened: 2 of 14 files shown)


thw26:
Another thought: just cross-linking an issue:

https://github.com/FaithLife-Community/OuDedetai/issues/194

thw26:
This PR is also necessary for #14

thw26:
Another note I made: https://github.com/FaithLife-Community/OuDedetai/issues/24#issuecomment-2224375729

What if we aimed for utilizing distrobox in order to simplify our whole process? We could potentially drop some of our dependency installations and rely (solely?) on distrobox.

We could then get around even the muslc (and FreeBSD?) issues by simply utilizing a Debian (or other: Arch to be most up-to-date?) system container.

Not having worked with distrobox, would we be able to prepare a container for distribution even? So instead of maintaining lengthy in-app instructions, could we distribute essentially our own Logos-focused Distro? i.e., maintain a distrobox container repo?

ctrlaltf24:
If we wanted a cross-distro solution flatpak/snap are more well traveled (and actually known). Regardless of our container solution, we'd still want wine as binaries we build

PR #437, opened 2025-08-29
Fix detect and avoid using chomium while broken

ctrlaltf24:
Don't love the user experience in this case: new users who happen to be using chromium based browser or snap or flatpak who don't care about the details of how OD works will get an additional prompt they don't really care about AND will be forced onto the beta channel of wine for future versions as well. Really as a user I'd almost just want to do this once (get the latest beta for this one release rather than all times), but that may not be a reasonable user experience. 

Due to the above concern this issue is still in draft. Don't want a patch today hurting users tomorrow.


ctrlaltf24:
Had a user test, worked with a snap browser

I tried to test with a snap browser and got it to launch firefox in a snap, but couldn't get back to OuDedetai (didn't detect the registration, and even if it did unsure if launching OD from INSIDE the snap is the best idea)

Ending confused, do we need to make something more elaborate with DBus messages? Would rather not, this is getting complicated...

PR #449, opened 2025-10-14
feat: display login workarounds in a pop up

ctrlaltf24:
Makes the user experience of browser login better by detecting likely failures and recommend solutions. Some problems still linger getting the browser to launch, then getting Logos to get the signal back.

https://github.com/FaithLife-Community/OuDedetai/issues/400
https://github.com/FaithLife-Community/OuDedetai/issues/435
https://github.com/FaithLife-Community/OuDedetai/issues/438

### DIFF
--- a/ou_dedetai/logos.py
+++ b/ou_dedetai/logos.py
@@ -117,7 +117,7 @@ class LogosManager:
         if get_logos_user_id(logos_appdata_dir) is not None:
             is_user_logged_in = True
 
-        sent_trouble_singing_in_message: bool = False
+        sent_trouble_signing_in_message: bool = False
 
         time_logos_started = time.time()
 
@@ -133,7 +133,7 @@ class LogosManager:
                     # Took me about ~1.5 minutes to sign on while simulating a little dilly-dallying and a couple
                     # incorrect password attempts. Should still be longer than most people take to sign on, but not SO
                     # long that people close OD in frustration to try again - fine line of balance.
-                    if not sent_trouble_singing_in_message and time.time() - time_logos_started > 60 * 1.75:
+                    if not sent_trouble_signing_in_message and time.time() - time_logos_started > 60 * 1.75:
                         # It's been five minutes since the user launched Logos and yet they haven't logged in yet.
                         # Are they having trouble?
                         self.app.pop_up(
@@ -149,7 +149,7 @@ class LogosManager:
                             f"If you still need help use \"Get Support\" on the main page of {constants.APP_NAME}"
                         )
                         # We only want to send this message once
-                        sent_trouble_singing_in_message = True
+                        sent_trouble_signing_in_message = True
 
 
     def start(self):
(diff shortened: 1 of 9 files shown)


ou_dedetai/logos.py
```diff
@@ -100,6 +101,57 @@ def monitor(self):
             # Useful if the install directory got deleted while executing
             self.logos_state = State.STOPPED
 
+    def monitor_for_potential_issues(self):
+        """A function to be run in parallel with running Logos to detect common issues and suggest workarounds"""
+        # Make sure we have a logos appdata, without it we can't reliably do any checks
+        logos_appdata_dir = self.app.conf._logos_appdata_dir
+        if logos_appdata_dir is None:
+            logging.warning("The user should have installed the Logos application by now and have this populated."
+                            " Waiting until that is resolved...")
```
thw26: nit: change "..." to "…"

ou_dedetai/logos.py
```diff
@@ -100,6 +101,57 @@ def monitor(self):
...
+            while logos_appdata_dir is None:
+                time.sleep(1)
+                logos_appdata_dir = self.app.conf._logos_appdata_dir
+        
+        # See if the user is logged in, this may change which things we look for
+        is_user_logged_in: bool = False
+        if get_logos_user_id(logos_appdata_dir) is not None:
+            is_user_logged_in = True
+
+        sent_trouble_singing_in_message: bool = False
```
thw26: Correct var name "sent_trouble_singing_in_message" to "sent_trouble_signing_in_message"

ou_dedetai/tui_curses.py
```diff
@@ -60,18 +60,18 @@ def text_centered(app: App, stdscr: curses.window, text: str, start_y=0) -> tupl
...
             write_line(
                 app,
                 stdscr,
                 text_start_y + i,
-                max(column_margin, x),
+                max(app.terminal_margin, x),
                 line,
                 # column_margin is doubled to account for the left side's column_margin
-                app.window_width - (column_margin * 2),
+                app.window_width - (app.terminal_margin * 2),
```
thw26, line 74: These changes don't quite equal what was present before.

column margin was double terminal margin, but here the values are effectively halved. terminal margin should be doubled in the first call and quadrupled in the second. No?

thw26 (changes requested):

ou_dedetai/tui_curses.py
ctrlaltf24: Correct, sorry didn't make a note of that. There was an issue with text centering function where if wrap was called on a line that was exactly terminal width - margin * 2 then the last two (aka terminal width) characters would be cut off and the left margin would be double the terminal width (4 instead of 2). Changes made restore terminal width to its proper one margin on each side (rather than the effective two on EACH side that was happening before)

thw26 (approved):

PR #451, opened 2025-10-14
feat(wip): change Logos-open-url to send messages over dbus

ctrlaltf24:
Created a tiny rust binary to send messages over dbus, then implemented a dbus server in python

Also refactored startup to initialize the python object then call start to enable being able to use the app instance in the dbus code.

TODO:
- [x] Still need to package and extract the binary
- [ ] double check dbus naming conventions to make sure the naming makes sense- much easier to change before we release anything
- [ ] test snap
- [ ] test flatpak
- [ ] test gui/tui exit conditions, need to make sure bases are covered after refactor
- [ ] test what happens when launching OD multiple times
- [ ] test opening link in the browser before OD starts - can we configure our .desktop files to open automatically?

Tested:
- install and run using chromium - works.


ou_dedetai/dbus.py
```diff
@@ -0,0 +1,25 @@
...
+from ou_dedetai.app import App
+from ou_dedetai.logos import LogosManager
+
+class FaithLifeAppInterface(ServiceInterface):
+    def __init__(self, app: App):
+        super().__init__("io.github.Faithlife_Community.OuDedetai.FaithLifeApp")
+        self.app = app
+
+    @method()
+    def Launch(self, args: 'as') -> None: # type: ignore[valid-type]
```
thw26, line 15: nit: comment should have two preceding spaces

thw26 (commented):
Did an initial code review. Still needs tested.

PR #456, opened 2025-12-11
gh pr checkout 389"ghcr.io/devcontainers-extra/features/amplify-cli:2": {}

carlajarchuleta1-a11y:


ctrlaltf24 (changes requested):
I'm very familiar with devconatiners and fine adding one here, but wouldn't we want the python one?

Issue #116, opened 2024-06-27
Implement a Logos Update Routine


thw26:
In order to get one of the most important features working, we need to implement an `update-logos()`. This function should accomplish the following steps:

1. Compare the user's installed Logos version to those available in the XML; if a new one is available, offer an update Logos button.
  - Near the top of `/home/user/LogosBible10/data/wine64_bottle/drive_c/users/deck/AppData/Local/Logos/System/LogosUpdater.deps.json`
2. Move the currently installed resources out of the way to be restored later using Linux `mv`, call it `move()`.
3. Install the newest MSI or a selected MSI, which due to a blank data folder wipes out the user's data folder.
4. Restore the user's data folder.
5. Run the `update-resources()`, in development by @n8marti.
6. Run the indexer.

This code may need to stop a running Logos client.

thw26:
Using the latest 36+, I was able to update books and my Logos install from within the app. Unfortunately, we are seeing the instant crash issue during this time, which is frustrating, but doesn't appear to be a complete showstopper.

I would think the code you have developed might still be useful, at least for partial updates.

ctrlaltf24:
Is this needed anymore? Logos updating itself is working right now

PR #473, opened 2026-06-16
feat: add "Update {product}" to re-run install at latest release

### LINKED ISSUE #272: Further consider upgrade behavior
Keeping in mind the priority not to break installations, how can we with a reasonable degree of safety roll out updates?

It would be nice if we could respect the user's choice on the automatic resource downloads setting.

For now in-app updates are disabled, obviously we cannot keep it like this. Either find a way to handle the update ourselves, or wait for Faithlife to hopefully apply the fix we need so we can let them do it again.

ctrlaltf24:
Add a button/menu/CLI entry that updates an installed Logos/Verbum to the latest available release by re-running the install routine.

The install chain is already idempotent except for ensure_product_installed, which skipped the MSI whenever the product exe existed regardless of version. Make that guard version-aware: run the MSI when the exe is missing, the installed release differs from the target, or the installed version can't be determined (deps.json missing/unreadable).

- utils.update_faithlife_product: pin release to the latest, then install()
- installer.ensure_product_installed: version-aware MSI guard
- gui/gui_app: "Update {product}" button wired to the new handler
- tui_app: "Update {product}" main-menu entry when installed
- cli/main: --update-installed-app subcommand (requires app installed)

Testing:
- Installed Logos 47.1, then hit "Update Logos" in the TUI. The main-menu version still showed 47, but after launching Logos and restarting OuDedetai it had updated to 51.1 (latest).
- Ran the update from the GUI while already on 51.1: install statuses cycled on the GUI, the process ran and cleared.
- Ran --update-installed-app from the CLI, same behavior as the GUI.

Fixes: https://github.com/FaithLife-Community/OuDedetai/issues/272 #116

### DIFF
--- a/ou_dedetai/gui.py
+++ b/ou_dedetai/gui.py
@@ -251,6 +251,10 @@ class ControlGui(StatusGui):
         # Install deps button
         self.deps_label = Label(self, text="Install dependencies")
         self.deps_button = Button(self, text="Install")
+        # Update Faithlife product button
+        self.update_product_labelvar = StringVar(value="Update Faithlife Product")
+        self.update_product_label = Label(self, textvariable=self.update_product_labelvar)
+        self.update_product_button = Button(self, text="Update")
         # Backup/restore data buttons
         self.backups_label = Label(self, text="Backup/restore data")
         self.backup_button = Button(self, text="Backup")
@@ -305,6 +309,9 @@ class ControlGui(StatusGui):
         row += 1
         self.deps_label.grid(column=0, row=row, sticky='w', pady=2)
         self.deps_button.grid(column=1, row=row, sticky='w', pady=2)
+        row += 1
+        self.update_product_label.grid(column=0, row=row, sticky='w', pady=2)
+        self.update_product_button.grid(column=1, row=row, sticky='w', pady=2)
         # row += 1
         # self.backups_label.grid(column=0, row=row, sticky='w', pady=2)
         # self.backup_button.grid(column=1, row=row, sticky='w', pady=2)
--- a/ou_dedetai/installer.py
+++ b/ou_dedetai/installer.py
@@ -372,7 +372,7 @@ def ensure_product_installed(app: App):
     try:
         installed_release = app.conf.installed_faithlife_product_release
     except Exception:
-        logging.warning("Failed to determine installed version; will install.", exc_info=True)
+        logging.warning("Failed to determine the installed version proceeding with install.", exc_info=True)
         installed_release = None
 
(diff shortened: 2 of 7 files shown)


ou_dedetai/gui.py
```diff
@@ -251,6 +251,10 @@ def __init__(self, root, *args, **kwargs):
         # Install deps button
         self.deps_label = Label(self, text="Install dependencies")
         self.deps_button = Button(self, text="Install")
+        # Update Faithlife product button
+        self.update_product_labelvar = StringVar(value="Update Faithlife Product")
```
thw26, line 255: Is it possible to use the actual installed product name here?

ou_dedetai/installer.py
```diff
@@ -369,13 +369,23 @@ def ensure_product_installed(app: App):
     app.installer_step += 1
     app.status(f"Ensuring {app.conf.faithlife_product} is installed…")
 
-    if not app.is_installed():
+    try:
+        installed_release = app.conf.installed_faithlife_product_release
+    except Exception:
+        logging.warning("Failed to determine installed version; will install.", exc_info=True)
```
thw26: Can we reword the "; will install." clause here? I'd rather this be two clear statements for if we ever implement localization.

ou_dedetai/utils.py
```diff
@@ -543,6 +543,18 @@ def update_to_latest_lli_release(app: App):
         logging.debug(f"{constants.APP_NAME} is at a newer version than the latest.") # noqa: 501
 
 
+def update_faithlife_product(app: App):
+    from ou_dedetai import installer  # local import to avoid circular dependency
+    releases = app.conf.faithlife_product_releases
+    if not releases:
+        app.status(f"No releases found for {app.conf.faithlife_product}; cannot update.")
```
thw26: Same comment about the semicolon clause.

thw26 (changes requested):

thw26:
The change on the unresolved comment I think still stands. I think there needs to be a period before "proceeding" and "proceeding" needs a capital P.

thw26 (changes requested):
Tested. Works! Woohoo!

Some issues found in TUI:

On updating, the TUI window does not change the Logos version until you relaunch.

If possible, we should offer an update for Logos only when it is available. Are we able to make that check and enable/disable the menu options appropriately?

ctrlaltf24:
Can we get this in and handle those in follow-on PRs?

thw26 (approved):

ctrlaltf24:
Not sure if we can hide that option without going to the internet. May not be worth doing. For our slow internet clients

PR #474, opened 2026-06-16
fix: restrict memory from halting system

### LINKED ISSUE #457: Limit Logos Memory Usage
Given recent version's of Logos using too much memory, establish the ability for OD to set the memory limit for the Logos process group.

This command was given as an example.

```
systemd-run --user --scope -p MemoryMax=3.5G -p MemoryHigh=2.5G /home/USER/.local/share/FaithLife-Community/oudedetai/oudedetai --run-installed-app
```

ctrlaltf24:
Two-layer memory protection prevents a runaway Wine/Logos process (e.g.
the indexer) from exhausting the host.

Primary – systemd-run cgroup enforcement:
When systemd-run is available, wine commands are wrapped in a transient
user scope with MemoryMax set to 90% of currently-available RAM. The
kernel enforces the limit at the cgroup level with no polling overhead.
Availability is probed once at runtime and cached.

Fallback – Python RSS watchdog:
A background thread (LogosManager._memory_watchdog) polls the resident
set size of the entire Wine process tree every two seconds. The cap is
90% of the RAM that would be free if the monitored process were not
running (available + process_rss), so the process's own growing
footprint is never counted against its budget — only other processes
can move the ceiling. If the cap is exceeded, the process group is
SIGTERM'd with a 10-second grace period before escalating to SIGKILL.

The watchdog starts when Logos launches or indexing begins, and is
stopped cleanly by end_processes(). Also fixes the AppImage download
gate to skip the download when the wine binary is set to a recommended
or beta sigil.

Tested:
- systemd-run cgroup path: Logos launched successfully inside a
  run-*.scope with MemoryMax=1612958515 (~90% of available RAM at
  launch). Scope appeared in `systemctl --user list-units --type=scope`.
- Fallback path: after removing the systemd-run binary, no scope was
  created and Wine launched unwrapped; watchdog remains the sole enforcer.
- Unit tests (13/13): all TestMemoryLimits and TestSystemdRunPrefix
  cases pass.
- Polling watchdog kill path: NOT yet tested end-to-end (no spike test
  run against a live process).
- Unittests on the polling watchdog as written

Have not tested with a real Logos installation - only mock tests. Mostly AI written tests. Would be wise to test manually if we know how to make the memory footprint trigger a crash

Fixes: #457

### DIFF
--- a/ou_dedetai/logos.py
+++ b/ou_dedetai/logos.py
@@ -267,7 +267,9 @@ class LogosManager:
         """Thread target: delegates to MemoryWatchdog and exits the app on a kill."""
         self._memory_watchdog.run()
         if self._memory_watchdog.kill_reason:
-            self.app.exit(self._memory_watchdog.kill_reason)
+            self.app.status(self._memory_watchdog.kill_reason)
+            # We don't want to kill the entire app here - CLI/TUI and GUI each handle when logos terminates differently.
+            # Existing code will already find out that Logos terminated and handle accordinly.
 
     def index(self):
         self.indexing_state = State.STARTING
--- a/ou_dedetai/logos.py
+++ b/ou_dedetai/logos.py
@@ -7,6 +7,7 @@ from enum import Enum
 import logging
 import psutil
 import threading
+from typing import Optional
 
 from ou_dedetai import database
 from ou_dedetai.app import App
@@ -32,6 +33,10 @@ class LogosManager:
         """These are sub-processes we started"""
         self.existing_processes: dict[str, list[psutil.Process]] = {}
         """These are processes we discovered already running"""
+        self._memory_watchdog_interval = 2
+        """Seconds between memory checks by the watchdog."""
+        self._memory_watchdog_stop = threading.Event()
+        self._memory_watchdog_thread: Optional[threading.Thread] = None
 
     def monitor_indexing(self):
         if self.app.conf.logos_indexer_exe_windows_path in self.existing_processes:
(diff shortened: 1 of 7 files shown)


ctrlaltf24:
The memory watchdog falls prey to https://github.com/FaithLife-Community/OuDedetai/issues/441 today. That issue will need to be fixed as well for this to be effective

ou_dedetai/logos.py
```diff
@@ -247,6 +253,22 @@ def end_processes(self):
...
+        ):
+            return
+        self._memory_watchdog.reset()
+        self._memory_watchdog_thread = self.app.start_thread(self._run_watchdog)
+
+    def _run_watchdog(self):
+        """Thread target: delegates to MemoryWatchdog and exits the app on a kill."""
+        self._memory_watchdog.run()
+        if self._memory_watchdog.kill_reason:
+            self.app.exit(self._memory_watchdog.kill_reason)
```
ctrlaltf24: Do we want this? Not sure that we do. If we're in the TUI for example we don't want the TUI to close just because Logos OOM'd. Fixing....

pydsigner:
> Would be wise to test manually if we know how to make the memory footprint trigger a crash

I was able to trigger a full system OOM freeze by opening enough Logos tabs to nearly fill system memory and then opening another Electron app that presumably triggered an overcommit. I've also been able to cause the OOM killer to terminate other apps by opening enough Logos resources. Do either of those cases sound like what you're seeking live testing for on this PR?

PR #475, opened 2026-06-16
chore: populate AI agent files

ctrlaltf24:
Using AI tools is assited by some context already being present so it doesn't need to scan the repo everytime. Saves AI uses (like myself) tokens


thw26:
I'm not familiar with these setups, but if you are good with this, I will approve for merge.

ctrlaltf24:
Yup. Feel free to read markdown if you want - but not required

thw26 (approved):

PR #480, opened 2026-06-20
fix: don't show update button when it isn't needed

ctrlaltf24:
Downside of this is it makes an additional network call on the main TUI page - which we had only one before for the other update button

On the GUI this network check is done on a separate thread in order to not halt the main body

Fixes: 478


thw26 (approved):

thw26:
Merging, but I will keep 478 open until the other aspects of the TUI window update properly.

thw26:
Since this is now merged, I will release a new beta so we can get the update button out.

PR #484, opened 2026-07-20
feat: keep track of network failures

ctrlaltf24:
A draft demonstrating how decorators can be used to facilitate not repeating the try except catches across the repo

@offline_aware guards NetworkRequests methods: a ConnectionError marks
the instance offline for 2 minutes and raises NetworkOffline. A
decorator wraps each method so the offline check and try/except frame
the body once, instead of repeating the same rule inside every method.

Ai generated untested


ctrlaltf24:
PR made for demonstration purposes more than an intent to merge. 

FYI @thw26

PR #487, opened 2026-08-24
fix: reject failed or incomplete downloads

sdwoodard:
## Summary

This PR fixes a download validation failure in which an HTTP error response,
interrupted transfer, or unverifiable cached file could be accepted as a valid artifact.

The fix is now contained within `logos_reuse_download` except for two commented cases
that could bypass it. Existing caller paths, destinations, and return-value behavior
remain unchanged. The narrowed behavior is covered by 30 deterministic tests with no
public network dependency.

## Root cause

The shared download path could write unsuccessful or interrupted responses into an
accepted location, while verification returned success when the server supplied neither
size nor checksum metadata. Resume could also append a response without first proving
that it was a valid continuation.

An existing recommended AppImage could skip the helper entirely, and self-update could
execute a cached updater before the helper validated it.

## Changes

* Require a final `200 OK`, finite `(10, 30)` timeouts, streaming, and a fresh GET from
  byte zero without `Range`.
* Treat HEAD only as optional cache-validation metadata. Without a trustworthy size or
  MD5, obtain a fresh response instead of accepting an existing file.
* Validate candidates from the established user download, application download, and
  target directories without changing those paths.
* Stage new responses and copies beside their destinations, then validate and publish
  with `os.replace()` while preserving existing files and permissions on failure.
* Accept a normally completed metadata-free `200` as transport complete, not
  cryptographically verified.
* Route existing recommended AppImages through the helper and validate cached
  self-updaters before executing them. Both locations include comments explaining why
  the exception is necessary.
* Replace the live `ip.me` test with mocks and reduce the suite to the narrowed behavior.

No dependency or CI configuration changes are included.

## Safety properties

After this change:

* unsuccessful or interrupted responses cannot replace accepted artifacts;
* unverifiable cached files are not reused;
* existing cache and target files survive failed replacement attempts; and
* failed downloads stop callers before copy, extraction, execution, installation,
  replacement, symlinking, or restart.

## Testing

Tested commit: `0537a3635ad20d85f2cd499c6326d037b8c03ae4`  
Base: `43fa3e508dce3f11a15339dd9e65a4973fc57ffc`

* `python -m unittest -v tests.test_network` — 30 of 30 passed.
* Python 3.12 with networking disabled — all 30 tests and Ruff passed, including a
  restrictive-umask run.
* Full-suite and mypy comparisons found no branch-only regression.
* A localhost truncated-transfer check preserved existing files and cleaned staging.
* `./scripts/build-binary.sh x86_64`, `--version`, and `--help` passed.
* The cumulative diff passes `git diff --check` and contains only `network.py` and
  `test_network.py`.

The full local verification run exits `1` because of four known upstream baseline
unittest failures on the host and 24 existing mypy findings. The focused tests, Ruff,
ancestry, diff, clean-tree, and unchanged-commit checks passed. The merge-base comparison
found no unittest or mypy regression unique to this branch.

## Compatibility and limitations

* Interrupted downloads restart from zero.
* Cache reuse requires trustworthy comparison metadata.
* A completed metadata-free `200` establishes transport completion, not cryptographic
  integrity.
* Existing caller paths and the implicit success return remain unchanged.


ctrlaltf24 (changes requested):
PR scope too large - rationale ambiguous.

What was the actual issue?

Why the sweeping changes?

sdwoodard:
Thanks for the feedback.  

This came from a static code scan which found that failed or interrupted downloads could be treated as valid when the server did not provide size or checksum metadata. That could allow an error response, partial transfer, or unverifiable cached file to continue into installation or execution. A failed replacement could also damage an existing valid file.  

The change reaches several callers because they all rely on the same download behavior and could continue after a failure. I addressed the shared failure contract rather than adding separate checks at each call site, and removed the unsafe resume path instead of adding more protocol complexity. Most of the added lines are tests covering those affected paths.  

I agree that the rationale should have been clearer and the review surface is large. I am happy to simplify or separate parts of the change if that would make it easier to review.

ctrlaltf24:
I'm mostly concerned with the things that appear to be changing paths and whatnot. If the changes could be contained to logos_reuse_download that would be ideal.

Sorry for coming on a little strong, but the testing required to validate this change is much higher than the slight benefit of having it. I agree for robustness this can be a good change - just trying to limit the fallout/impact.

sdwoodard:
Ok, that's reasonable. I can rework this so the changes are contained within logos_reuse_download.  

There are two calls that would remain outside that protection. One skips logos_reuse_download when the recommended AppImage already exists. The other executes a cached updater binary before logos_reuse_download is called. Would you prefer those remain as they are or should those be updated?

They're uncommon cases, but an interrupted or corrupted cached file could potentially be used before any validation runs.

ctrlaltf24:
I don't see the changes you're speaking about in the diff. 

How about this: anything outside of logos_reuse_download have a comment rationalizing the change and try to keep the changes outside that scope to a minimum. Sound reasonable?

sdwoodard:
Yup, thats fair. I reworked this so most of the changes are limited to `network.py` as requested.  

Outside `logos_reuse_download`, only the two cases mentioned earlier remain and they have comments for the rationale.  

The PR body has also been updated with the details for the new changeset.

ctrlaltf24:
PR is still 930 lines long - that's really long for this scope of change.

Issue #247, opened 2025-01-13
Resizing Less than Perfect in Beta 4


thw26:
The resizing of the TUI is still less than perfect.

During the install, Logos had finished installing and the final window was popped up asking me to "Finish". I didn't click this but instead resized my terminal (using KDE's Konsole).

I then clicked finish and the screen updated to ask me to write the config file. I clicked okay but the screen didn't update. I dragged the screen to force a refresh and it resized and updated the window.

The console log continues to update (hitting bad keys to see "Input Unknown") but the menu itself did not update its highlighted option. Resizing doesn't appear to be working now and the menu items are not updating.

I'm sure once I exit and restart OD that all will be well.

thw26:
#304 resolved one of the main issues.

Resolving #248 and #303 would likely resolve most other issues.

PR #491, opened 2026-08-30
fix: don't call endwin() from the SIGINT handler

### LINKED ISSUE #489: Ctrl+C can exit with a traceback and leave the terminal unrestored: endwin() runs in the SIGINT handler

jbellars:
Fixes #489.

`TUI.end` is installed as the SIGINT handler and tears curses down inline:

```python
def end(self, signal, frame):
    logging.debug("Exiting…")
    self.is_running = False
    curses.endwin()
```

SIGINT can land anywhere — in the reported traceback, inside `stdscr.getch()`. `endwin()` runs, but the display loop is still mid-iteration and calls `refresh()` before it next checks `is_running`, which puts the screen back in running state. `curses.wrapper()`'s teardown `endwin()` then returns ERR, and because that fires from the `finally`, it replaces the clean exit with an unhandled exception and the terminal is never restored:

```
  File "ou_dedetai/tui_app.py", line 410, in end
_curses.error: endwin() returned ERR
```

## The change

The `endwin()` call is redundant as well as racy. `App.exit()` ends in `sys.exit()`, which propagates through `display()` and `run()`'s `finally` into `curses.wrapper()`, and the wrapper tears curses down on the way out — there is no `os._exit()` anywhere in the package, so that path is never bypassed. The handler now only stops the loop.

## Verification

Sending Ctrl+C through a pty once the control panel has settled:

| build | dirty |
|---|---|
| released 4.0.0-beta.14 binary (py3.12) | 0/5 |
| locally built, same source, py3.14 | 4/5 |
| this branch, py3.14 | 0/8 |

Worth being straightforward about: I could not reproduce this on the official 3.12 release binary, and CONTRIBUTING.md is explicit that the app is built with 3.12 and may misbehave on other versions. But one of the 4/5 runs came out clean, which reads as a race that different build timings win or lose rather than a 3.14-specific fault — the 3.12 build wins it, it doesn't avoid it. #247 reports the TUI being left needing a `reset` after exiting, and a failed `endwin()` is one way to land exactly there.

Resize behaviour is unchanged (verified separately at 24x80, 50x160 and 33x100). `ruff check` clean; `mypy ou_dedetai` reports nothing new; unit tests go 77 → 78 with the same 3 failures + 1 error already present on `main`.

## Note on the test file

`tests/test_tui_app.py` is new, and the sibling PR for #488 also creates it with its own tests. Whichever lands second will need a trivial rebase — keep both classes. Happy to restructure or fold them together if you'd rather take them as one.


thw26:
Sorry we have been delayed on reviewing this. I will talk with @ctrlaltf24 to see if he is available to do so.

ou_dedetai/tui_app.py
```diff
@@ -403,9 +403,13 @@ def _exit(self, reason, intended = False):
         self.end(None, None)
 
     def end(self, signal, frame):
+        # Only ask the loop to stop. This runs as the SIGINT handler, so tearing
+        # curses down here races the loop's next refresh(), which puts the screen
+        # back in running state; the teardown endwin() then fails with
+        # "endwin() returned ERR" and the terminal is left unrestored. run()'s
+        # finally and curses.wrapper() already do the teardown on the way out.
```
ctrlaltf24: Remove this comment entirely. Claude code is prone to meta-commentate. That is say why it did it in a comment. Normally it's too verbose. Here in this case this is what the git log is for. These comments are only occasionally useful (and always verbose). The phrase I use after Claude written code is "trim all phrases that don't bear weight". Helps

ctrlaltf24 (approved):

ou_dedetai/tui_app.py
jbellars: Done — removed. Point taken on the meta-commentary; the reasoning lives in the commit message, which is the right place for it.

I applied the same trim to the one other spot it had crept into: the class docstring in `tests/test_tui_app.py` was restating the same race in five lines, now just `"""SIGINT must only stop the loop."""`.

Squashed into the single fix commit and force-pushed (`894d72b` → `2e10d6d`), so `tui_app.py`'s whole contribution to this PR is now the one `-curses.endwin()` line.

jbellars:
Updated. Thanks for the feedback!

PR #495, opened 2026-09-14
Switch TUI Library from Curses to Textual

### LINKED ISSUE #228: Switch TUI to urwid

### LINKED ISSUE #477: TUI Not Properly Updating After Change
On Logos update or release channel change, the TUI window doesn't update its identifiers.

### LINKED ISSUE #488: TUI crashes with RecursionError: SIGWINCH handler redraws inline and re-enters itself

### LINKED ISSUE #489: Ctrl+C can exit with a traceback and leave the terminal unrestored: endwin() runs in the SIGINT handler

### LINKED ISSUE #493: DXVK makes the Logos 10 UI responsive; could it be a renderer option?
With the default `renderer=gdi`, Logos draws everything on the CPU, and scrolling and typing lag (the same symptoms as #444 and #79). We installed DXVK into the prefix so Logos draws through Vulkan on the GPU instead. On both machines we tried, Logos went from laggy to usable. The GDI to GL switch in #444 sped things up but brought flicker and black windows. We haven't seen that with DXVK.

Tested on:

- Desktop: Fedora KDE, Radeon RX 7700 XT (Mesa RADV), Logos 10 53.1.0.0002, fresh install
- Laptop: Fedora 44, Intel Raptor Lake-U integrated graphics (Mesa ANV), existing install updated in place
- Both: Ou Dedetai v4.0.0-beta.14, wine-staging 10.8 AppImage (wow64), DXVK 3.1 installed with winetricks

Steps, with Logos closed (adjust the path if your `install_dir` in `~/.config/FaithLife-Community/oudedetai.json` is different):

```bash
cd ~/.local/share/FaithLife-Community/oudedetai/data
export WINEPREFIX=$PWD/wine64_bottle WINE=$PWD/bin/wine WINESERVER=$PWD/bin/wineserver
./bin/wineserver -w
cp $WINEPREFIX/user.reg $WINEPREFIX/user.reg.pre-dxvk
cp $WINEPREFIX/system.reg $WINEPREFIX/system.reg.pre-dxvk
./bin/winetricks -q dxvk
./bin/wine reg delete 'HKCU\Software\Wine\Direct3D' /v renderer /f
```

Leave "Enable Hardware Acceleration" in Logos set to Yes (the default). To check that DXVK loaded, launch with `DXVK_HUD=devinfo` and the overlay shows your GPU and driver. We haven't needed to revert, but copying the two `.pre-dxvk` files back and running `wine wineboot -u` should undo it.

Scrolling on the desktop went from roughly 5 to 15 fps with stutter to smooth, and the laptop went from unusable to usable. We also took 10-second `top` samples of Logos.exe while scrolling: desktop 104% down to 84-97%, laptop 137% down to 100%. Those numbers swing a lot with scroll speed, so the change in feel is the better evidence. Memory use went up about 0.5 GB on the desktop and didn't change on the laptop.

The DXVK FPS counter isn't much use with Logos. Logos only presents a frame when something changes, so the number freezes between redraws. Also, the first time you hover a word, the Information tooltip leaves a black strip at the end of the bubble, and hovering again clears it. It may have been there before DXVK. Turning off "Show Information Tooltips" gets rid of it.

Suggestion: add `dxvk` to the Wine Config renderer list next to gdi, gl and vulkan, keep gdi as the default, and point slow-UI reports to it.

thw26:
This PR is intended to supersede all Curses code by swapping the library to Textual. This gives us a lot of benefits over our existing code, such as better styling features. Not the least of which is a reduction of a net of nearly 2,000 lines.

Textual also is/can be touch-screen friendly. I have used it in another application I have developed and I find it much nicer to work with than Curses.

This PR fixes #228 and thus also supersedes #490 and #491. It would thereby also close without needing a fix #488 and #489.

This also fixes #477, #141, and #247.


ou_dedetai/installer.py
```diff
@@ -332,9 +332,9 @@ def ensure_wineprefix_config(app: App):
     logging.debug("Setting wineprefix registry to ignore winemenubuilder.exe.")
     wine.disable_winemenubuilder(app=app, wine64_binary=app.conf.wine64_binary)
 
-    # Force renderer=gdi in registry.
-    logging.debug("Setting renderer=gdi in wineprefix registry.")
-    wine.set_renderer(app=app, wine64_binary=app.conf.wine64_binary, value='gdi')
+    # Force renderer=dxvk in registry.
+    logging.debug("Setting renderer=dxvk in wineprefix registry.")
```
ctrlaltf24, line 336: Separate PR please - I want to see this commit in the git log

ou_dedetai/tui_app.py
```diff
@@ -1,1358 +1,658 @@
...
-from . import msg
 from . import system
-from . import tui_curses
-from . import tui_screen
 from . import utils
 from . import wine
 
-console_message = ""
 
+WINDOWS_VERSIONS = ["vista", "win7", "win8", "win10", "win11", "Return to Main Menu"]
```
ctrlaltf24, line 31: This doesn't belong here

ou_dedetai/tui_app.py
```diff
@@ -1,1358 +1,658 @@
...
+            primary="#00AEEF",
+            secondary="#666666",
+            accent="#FFFFFF",
+            background="#101010",
+            surface="#181818",
+            panel="#202020",
+            foreground="#F0F0F0",
+            dark=True,
+        ),
+        #TODO: Needs tweaking
```
ctrlaltf24, line 124: @thw26

ou_dedetai/tui_app.py
```diff
@@ -1,1358 +1,658 @@
...
+            warning="#F2C14E",
+            error="#E06C75",
+            success="#98C379",
+            foreground="#E6E6E6",
+            background="#121212",
+            surface="#1E1E1E",
+            panel="#252525",
+            dark=True,
+        ),
+        #TODO: Needs tweaking
```
ctrlaltf24, line 139: @thw26

ou_dedetai/tui_app.py
```diff
@@ -1,1358 +1,658 @@
...
+            error="#B3261E",
+            success="#137333",
+            foreground="#202124",
+            background="#FFFFFF",
+            surface="#F5F5F5",
+            panel="#E8E8E8",
+            dark=False,
+        )
+    }
+    #TODO: Needs tweaking
```
ctrlaltf24, line 155: @thw26

ou_dedetai/tui_app.py
```diff
@@ -1,1358 +1,658 @@
...
+
+    #dialog {
+        width: 70;
+        max-width: 90%;
+        height: auto;
+        border: thick $primary;
+        background: $surface;
+        padding: 1 2;
+    }
+
```
ctrlaltf24, line 203: I wonder how much of this custom css is actually needed

ou_dedetai/tui_app.py
```diff
@@ -1,1358 +1,658 @@
...
+        background: $panel;
+        color: $text-muted;
+        padding: 0 1;
+    }
+    """
+
+    BINDINGS = [
+        Binding("q", "quit", "Quit", show=True),
+        Binding("ctrl+c", "quit", "Quit", show=False),
+        Binding("r", "refresh", "Refresh", show=True),
```
ctrlaltf24, line 234: Refresh? Like force a screen refresh?

ou_dedetai/tui_app.py
```diff
@@ -1,1358 +1,658 @@
...
+    def compose(self) -> ComposeResult:
+        with Vertical(id="header-area"):
+            yield Static(id="title")
+            yield Static(id="subtitle")
+        yield Log(id="console-area", max_lines=300, highlight=True)
+        with Vertical(id="main-area"):
+            yield Static("Main Menu", id="menu-title")
+            yield OptionList(id="main-menu")
+        yield Footer()
+        yield Static("", id="status-bar")
```
ctrlaltf24, line 257: That's...a really strange programming pattern. It's like composing a list by yielding. I'd flag this, this should be changed unless it fits textual's stuff. It's difficult to understand/wrap your head around proses like this. Few people understand yield semantics- like did you know the function stack for yield is actually stored? Stuff is kept in memory more than you need for a list return

ou_dedetai/tui_app.py
```diff
@@ -1,1358 +1,658 @@
...
+from textual.app import App as TextualApp, ComposeResult
+from textual.binding import Binding
+from textual.containers import Container, Horizontal, Vertical
+from textual.reactive import reactive
+from textual.screen import ModalScreen
+from textual.theme import Theme
+from textual.widgets import (Button, Footer, Input, Label, Log, OptionList, Static)
+from textual.widgets.option_list import Option
+
+from ou_dedetai.app import App as DomainApp, UserExitedFromAsk
```
ctrlaltf24, line 19: I would prefer the inputs renamed the other direction. Textual being TextualApp and OD being either raw app or OuDedetaiApp

ou_dedetai/tui_app.py
thw26: This is the standard way that Textual is programmed. It's the one oddity that was goofy having to learn.

ou_dedetai/installer.py
thw26: Got it. Will do.

ou_dedetai/tui_app.py
thw26: Will move.

ou_dedetai/tui_app.py
thw26: I'll test with it removed and add back in what is needed to see. Some of it is necessary.

ou_dedetai/tui_app.py
thw26: Got it. Will refactor.

ou_dedetai/tui_app.py
thw26: In Curses we had the ability to change color scheme, in case someone found our default unreadable. Textual doesn't have a system theme like Curses did, so I was attempting to implement some defaults.

ou_dedetai/tui_app.py
thw26: In Curses we had the ability to change color scheme, in case someone found our default unreadable. Textual doesn't have a system theme like Curses did, so I was attempting to implement some defaults.

ou_dedetai/tui_app.py
thw26: I think I can remove the refresh. Will test.

ou_dedetai/tui_app.py
thw26: I minimized the CSS in the most recent commit. The CSS that is there either brings unity to the design or readability.

PR #496, opened 2026-09-16
fix: #493, set default renderer to dxvk

### LINKED ISSUE #79: Slow Non-Responsive UI
Suddendly on my side the isntalled app with alpha 3 is not working properly anymore. I can't really use it due it really laggy. If i try to enter text for example as note or in the sermon editor it takse a few seconds until input appears. 

I tried to disable hardware acceleration and also reinstalled everything. 

What could causing that issue?

### LINKED ISSUE #444: Logos UI slow when on a 4k screen
When I use Logos on a resolution of 3840x2160 the UI gets really laggy, but when I use a lower resolution like 1920x1080 it works fine. 
For the record I use a HP Spectre with a GeForce GTX 1650,  Intel i7-9750H and 16 gb ram so the hardware is not the issue, more like the UI isn't configured for a 4k screen at least on my computer.

### LINKED ISSUE #493: DXVK makes the Logos 10 UI responsive; could it be a renderer option?

thw26:
Fixes #493 by setting the default renderer to DXVK. It thus may also fix #79 and #444, once people run either the set renderer command or install again.


ctrlaltf24:
Let's wait for @jg00dman on this one due to 

> But I found that now logos gets mouse clicks and scroll interaction even when below another window!

https://github.com/FaithLife-Community/OuDedetai/issues/493#issuecomment-5665515962

PR #497, opened 2026-09-19
fix: black frames around Wine popups on Hyprland/Sway (XShape shim)

mddavilap:
## Summary

On Hyprland (and other wlroots-style compositors such as Sway), Logos under OuDedetai shows an opaque **black frame** around every popup with a drop shadow: word-info tooltips, the Library popup, the Library right-click menu, study-Bible reference popovers, plus a black flash when dialogs close. The same install is fine on GNOME and KDE. #493 mentions a "black strip" on tooltips in passing — same mechanism.

This PR adds a tiny `LD_PRELOAD` shim that removes the frames, enabled automatically on Wayland sessions that are not GNOME/KDE.

> **Disclosure — please review accordingly:** the diagnosis below and all of the code in this PR were produced by Claude (Anthropic's AI model), working interactively with me on my Omarchy/Hyprland laptop. I'm the user who hit the bug and tested the fix; I did not write the analysis or the code myself. I'm flagging this so you can scrutinise the reasoning and the code with the appropriate level of care rather than assume a human maintainer's familiarity with Wine/X11 internals is behind it. Everything stated below was verified on my machine and the reproduction program is included so you can check it independently.

## Root cause

1. Since Wine 9.12 (commit `0931c2a4`, *"win32u: Update the window surface shape with color key and alpha"*), every per-pixel-alpha layered window gets an **XShape bounding region derived from its alpha channel** (`dlls/win32u/dce.c: shape_from_alpha_mask`, applied in `dlls/winex11.drv/bitblt.c: x11drv_surface_flush` through `XShapeCombineMask`). The outline of that shape is where the popup's soft shadow fades to alpha 0 — i.e. exactly the rounded rectangle you see as the black frame.
2. Hyprland does not apply XShape when compositing XWayland surfaces, and XWayland leaves the pixels outside the shape stale in the buffers it hands to the compositor. So the transparent shadow margin renders as opaque black. Mutter and KWin apply the shape as a compositing mask, hence "works on GNOME/KDE".

Reproduced independently of Logos with the program at the end of this description: an ARGB override-redirect window renders correctly under Hyprland; add a bounding shape via `XShapeCombineMask` and every alpha-0 pixel turns black. Reading the window's pixels back from the X server (`XGetImage`) shows correct alpha in both cases — what Wine draws is right; the shape is what breaks compositing.

(A second, DXVK-only variant exists: with `winetricks dxvk`, `winex11.drv/vulkan.c` creates a depth-24 client child window inside the ARGB popup and XWayland's "surface window" logic exports that opaque child instead of the parent. That only appears with DXVK; the XShape one is present with the default GDI renderer. Not addressed here.)

## The fix

Under a compositor the alpha mask serves no purpose (it exists so non-compositing X setups can click through transparent areas). The shim interposes `XShapeCombineMask` and drops only calls with `ShapeBounding` and a real mask pixmap. Shape resets (`src == None`) and `XShapeCombineRectangles`/`Region` — which Wine uses for `SetWindowRgn()` and to hide zero-sized windows — pass through untouched.

- `ou_dedetai/assets/nowineshape.c` — the shim, ~15 lines of C, no X11 headers needed (only libc/libdl), so it builds anywhere with a C compiler.
- `ou_dedetai/xshape_shim.py` — `wanted()`: `XDG_SESSION_TYPE=wayland` and `XDG_CURRENT_DESKTOP` not containing GNOME/KDE, overridable with `OUDEDETAI_XSHAPE_SHIM=1/0`. `library_path()`: builds the `.so` once with `$CC`/`cc`/`gcc`/`clang` into `<install_dir>/data/`, rebuilds if the source is newer, logs a warning once and does nothing if there is no compiler or the build fails. `apply()`: prepends it to `LD_PRELOAD`.
- `ou_dedetai/wine.py` — one call in `get_wine_env()`, and `LD_PRELOAD` added to the debug log line.

The shim is applied to every Wine process we spawn (winetricks, wineserver, etc.), which is harmless — it only affects the one X call.

## Testing

- Hyprland 0.56.2, xorg-xwayland 24.1.13, Omarchy (Arch), hybrid Intel/NVIDIA laptop; OuDedetai 4.0.0-beta.14 + wine-staging 10.8 AppImage, Logos 53.1.0.0002, default `renderer=gdi`.
- Ran from this branch (`python -m ou_dedetai.main --run-installed-app`): `nowineshape.so` built with `cc` into `data/`, `LD_PRELOAD` present in the Wine env, shim mapped in `Logos.exe` (`/proc/<pid>/maps`), and a full session with no black frames: tooltips, Library popup, context menus, reference popovers, dialog close animation. No crashes or regressions noticed.
- `wanted()` checked for Hyprland/GNOME/KDE/X11 and both override values.
- `ruff check` clean; `mypy` reports no new errors (the 15 existing ones are in unrelated files).

## Things you may want to decide

- Whether you'd rather expose this as a config option than an env override.
- Whether to ship a prebuilt `.so` from CI instead of compiling on the user's machine (a C compiler is a new soft dependency; the fallback is just a logged warning).
- Long term this belongs upstream — either Hyprland honouring XShape on XWayland surfaces, or Wine skipping the alpha-derived shape under a compositor. Happy to open those reports too.

## Reproduction program (no Wine or Logos needed)

```c
/* gcc argbshape.c -o argbshape -lX11 -lXext ; run under Hyprland over a light background */
#include <X11/Xlib.h>
#include <X11/Xutil.h>
#include <X11/extensions/shape.h>
#include <stdlib.h>
#include <unistd.h>
int main(int argc, char **argv) {
    Display *d = XOpenDisplay(NULL); int s = DefaultScreen(d); XVisualInfo vi;
    XMatchVisualInfo(d, s, 32, TrueColor, &vi);
    XSetWindowAttributes a = {0}; a.colormap = XCreateColormap(d, RootWindow(d, s), vi.visual, AllocNone);
    a.override_redirect = True; int W = 400, H = 200;
    Window w = XCreateWindow(d, RootWindow(d, s), 900, 700, W, H, 0, 32, InputOutput, vi.visual,
                             CWColormap | CWBorderPixel | CWBackPixel | CWOverrideRedirect, &a);
    if (argc > 1) { /* "shape": bounding mask like Wine's alpha>0 region */
        Pixmap m = XCreatePixmap(d, w, W, H, 1); GC g = XCreateGC(d, m, 0, NULL);
        XSetForeground(d, g, 0); XFillRectangle(d, m, g, 0, 0, W, H);
        XSetForeground(d, g, 1); XFillRectangle(d, m, g, 20, 20, W - 40, H - 40);
        XShapeCombineMask(d, w, ShapeBounding, 0, 0, m, ShapeSet);
    }
    XMapWindow(d, w);
    XImage *img = XCreateImage(d, vi.visual, 32, ZPixmap, 0, malloc(W * H * 4), W, H, 32, 0);
    for (int y = 0; y < H; y++) for (int x = 0; x < W; x++) {
        unsigned p = 0;                                   /* transparent margin      */
        if (x >= 40 && x < W - 40 && y >= 40 && y < H - 40) p = 0xFFFF0000; /* opaque red */
        XPutPixel(img, x, y, p);
    }
    GC gc = XCreateGC(d, w, 0, NULL);
    for (int i = 0; i < 16; i++) { XPutImage(d, w, gc, img, 0, 0, 0, 0, W, H); XFlush(d); usleep(500000); }
    return 0;
}
```

`./argbshape` → red rectangle with a see-through margin. `./argbshape shape` → the margin is solid black. `LD_PRELOAD=./nowineshape.so ./argbshape shape` → see-through again.


thw26:
cf. #496

ctrlaltf24:
Admittedly a pretty cool fix

But I can't speak much to the robustness of it. Reimplementing a single function in C seems...risky or otherwise a last resort. 

If switching the renderer fixes this I'd prefer that. Does the PR that T.H. Linked resolve your issue as well?

ctrlaltf24:
On second thought this might be useful to open up against upstream   (Rather than patching it like this here). Patching one function out of a binary is...risky or fragile. I've definitely done it before as a last resort but the upstream function changes this could introduce a bug/difference with upstream wine.

Better that everyone gets this fix

We have LONG considered bundling/patching our own wine and are close/far. Only a couple iterations left to get that working. Don't have the time this second to do it though

